PR #23, opened 2014-11-03
fix for wrong tree structure upon importing an qti package.

aamirr:
http://forge.taotesting.com/issues/2993

### DIFF
--- a/model/qti/ImportService.php
+++ b/model/qti/ImportService.php
@@ -31,6 +31,10 @@ use \taoItems_models_classes_ItemsService;
 use \common_exception_Error;
 use \common_ext_ExtensionsManager;
 use \core_kernel_classes_Property;
+use \core_kernel_classes_Session;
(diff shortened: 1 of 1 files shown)


model/qti/ImportService.php
```diff
@@ -117,7 +121,26 @@ public function importQTIFile($qtiFile, core_kernel_classes_Class $itemClass, $v
             
             //save itemcontent
             if($qtiService->saveDataItemToRdfItem($qtiItem, $rdfItem)){
+
```
ssipasseuth: It might work but I don't think that this piece of code should be located here. 
Shouldn't this fix be part of oat\taoQtiItem\model\qti\ManifestParserFactory::getResourcesFromManifest() called in  oat\taoQtiItem\model\qti\ImportService::importQTIPACKFile() instead ?

ssipasseuth:
This branch is outdated.

PR #44, opened 2015-01-08
modify alt text in qti item view

antoinerobin:
https://github.com/oat-sa/tao-core/pull/94


views/js/qtiCreator/widgets/static/img/states/Active.js
```diff
@@ -232,10 +233,32 @@ define([
...
-                                img.attr('alt', label);
-                                $label.val(label).trigger('change');
+                            if($.trim($alt.val()) === ''){
+                                if(alt === ''){
+                                    alt = _extractLabel(file);
+                                }
+                                img.attr('alt', alt);
+                                $alt.val(alt).trigger('change');
+                            }
+                            else if(alt !== ''){
```
ssipasseuth: I wonder if the "else" condition is actually relevant ... have you got a reasonable explication for this ?

PR #47, opened 2015-01-09
Integration portable shared aliases

ssipasseuth:


ssipasseuth:
WIth this merged, we can finally delete the tmp branch !

PR #51, opened 2015-01-16
Itempack

krampstudio:
Implement Packing (alternative compilation) for QTI Items.
Requires https://github.com/oat-sa/extension-tao-item/pull/28


krampstudio:
@ssipasseuth should have also a look (unable to assign 2 people on a PR)

PR #52, opened 2015-01-16
remove local constant from unit tests

antoinerobin:


jbout:
Introduces a new failure:
Failed asserting that exception of type "common_exception_Error" matches expected exception "oat\taoQtiItem\model\qti\exception\ParsingException".

PR #58, opened 2015-01-21
Feature/extended textinteraction counter

ssipasseuth:


ssipasseuth:
Merge for sprint ending. Only format field can be edited properly so far.

PR #59, opened 2015-01-22
load image on item load

antoinerobin:
load them even with a link with #, ? ... (due to new filemanager implementation in mediamanager)

### DIFF
--- a/views/js/qtiCreator/widgets/static/img/Widget.js
+++ b/views/js/qtiCreator/widgets/static/img/Widget.js
@@ -25,7 +25,7 @@ define([
         //check file exists:
         inlineHelper.checkFileExists(this, 'src', options.baseUrl);
         $('#item-editor-scope').on('filedelete.resourcemgr.' + this.element.serial, function(e, src){
-            if(itemUtil.fullpath(img.attr('src'), baseUrl) === itemUtil.fullpath(src, baseUrl)){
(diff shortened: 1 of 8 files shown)


views/js/qtiCreator/widgets/static/img/Widget.js
```diff
@@ -25,7 +25,7 @@ define([
         //check file exists:
         inlineHelper.checkFileExists(this, 'src', options.baseUrl);
         $('#item-editor-scope').on('filedelete.resourcemgr.' + this.element.serial, function(e, src){
-            if(itemUtil.fullpath(img.attr('src'), baseUrl) === itemUtil.fullpath(src, baseUrl)){
+            if(itemUtil.fullpath(encodeURIComponent(img.attr('src')), baseUrl) === itemUtil.fullpath(encodeURIComponent(src), baseUrl)){
```
krampstudio: Be careful with `encodeURIComponent` : 
```
encodeURIComponent('http://foo.bar');
//returns 'http%3A%2F%2Ffoo.bar'
```

krampstudio:
I need more time to review it in details.

krampstudio:
I think those changes aren't accurate. If the resource URL contains parameters this should be encoded, I think the server API must return the URL encoded  correctly. The client shouldn't have to take care of that.

PR #71, opened 2015-02-06
allow saving only actual number value

ssipasseuth:


ssipasseuth:
somehow related to https://github.com/oat-sa/tao-core/pull/126

PR #80, opened 2015-02-09
Operator lt

boajer:


krampstudio:
be careful, it seems you mess up subtract.js with lt.js

views/js/scoring/processor/expressions/operators/subtract.js
```diff
@@ -34,7 +34,7 @@ define([
      * @type {OperatorProcesssor}
      * @exports taoQtiItem/scoring/processor/expressions/operators/subtract
      */
-    var subtractProcessor = {
+    var ltProcessor = {
```
krampstudio: There is still those changes that should be reverted.

PR #81, opened 2015-02-10
introduction of gte operator

boajer:


views/js/scoring/processor/expressions/operators/gte.js
```diff
@@ -0,0 +1,76 @@
...
+            if(_.some(this.operands, _.isNull) === true){
+                return null;
+            }
+
+
+            result.value = preProcessor
+                .mapNumbers(this.operands)
+                .reduce(function(f, s){
+                    return f >= s;
+                });
```
krampstudio: could you run somes tests and send me the results of a console.log in the reduce function:

```javascript
.reduce(function(f, s){
    console.log(f + " >= " + s,   f >= s);
    return f >= s;
});
```

boajer: gte.js:69 5 >= 2 true
gte.js:69 5 >= 5 true
gte.js:69 15 >= 5 true
2gte.js:69 0.666677 >= 1.333323 false
gte.js:69 5 >= 10.25 false

krampstudio: Ok cool. 
I need to ensure the number of operands on a higher level

PR #82, opened 2015-02-10
Fix modal feedback

ssipasseuth:
fixed http://forge.taotesting.com/issues/3295 and http://forge.taotesting.com/issues/3298

Their is still the z-index bug on authoring form. This will be addressed in another pull request


ssipasseuth:
this is requires https://github.com/oat-sa/tao-core/pull/129 and https://github.com/oat-sa/tao-core/pull/128

PR #89, opened 2015-02-13
Feature rp parsing json

ssipasseuth:
Add parsed responseRules into the output of the responseProcessing::toArray()


krampstudio:
seems good.
I'll test it more carefully

PR #90, opened 2015-02-13
Truncate operator

boajer:


views/js/scoring/processor/expressions/operators/truncate.js
```diff
@@ -0,0 +1,74 @@
...
+            }
+
+            var value = preProcessor
+                .parseOperands(this.operands).value()[0];
+
+            if ( _.isNaN(value) ) {
+                return null;
+            }
+
+            result.value = _.isFinite(value) ? Math.trunc(value) : value;
```
krampstudio: Be careful `Math.trunc` is part of ES6, you should'nt use it (at least right now :) Please see https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Math/trunc

PR #91, opened 2015-02-13
Operator round to

boajer:


views/js/scoring/processor/expressions/operators/roundTo.js
```diff
@@ -0,0 +1,128 @@
...
+            var result = {
+                cardinality : 'single',
+                baseType    : 'float'
+            };
+
+            //if at least one operand is null, then break and return null
+            if(_.some(this.operands, _.isNull) === true){
+                return null;
+            }
+            var value = preProcessor
```
krampstudio: You can use `preProcessor.parseVariable` instead when min and max are set to 1

views/js/scoring/processor/expressions/operators/roundTo.js
```diff
@@ -0,0 +1,128 @@
...
+    'use strict';
+
+    /**
+     * Process operands and returns roundTo result.
+     * @type {OperatorProcessor}
+     * @exports taoQtiItem/scoring/processor/expressions/operators/roundTo
+     */
+    var roundToProcessor = {
+
+        engines : {
```
krampstudio: Why don't you just add methods with the name of the engine instead of having a string that gives you the function to execute ?

views/js/scoring/processor/expressions/operators/roundTo.js
```diff
@@ -0,0 +1,126 @@
...
+        operands   : [],
+
+        /**
+         * @returns {?ProcessingValue} a single boolean
+         */
+        process : function(){
+
+            var roundingMode  = this.activeRoundingEngine;
+            var figures = this.figures;
+
```
krampstudio: You can get attribute values from `this.expression` ie. `this.expression.figures`  also the attribute is `roundingMode` and not `activeRoundingEngine`

views/js/scoring/processor/expressions/operators/roundTo.js
```diff
@@ -0,0 +1,121 @@
...
+        operands   : [],
+
+        /**
+         * @returns {?ProcessingValue} a single boolean
+         */
+        process : function(){
+
+             var roundingMode = _.isFunction(this.expression.attributes.roundingMode) ? this.expression.attributes.roundingMode : roundToProcessor.engines.significantFigures;
+            var figures = _.isNumber(this.expression.attributes.figures) ? this.expression.attributes.figures : undefined;
+
```
krampstudio: this.expression.attributes are directly retrieved from the QTI XML. 
`this.expression.attributes.roundingMode` will be a string and `this.expression.attributes.figures` will be either a string or number that represents a string (use parseInt) or the reference to a variable, then you have to get the variable from `this.state` and get it's value, like (but implement it more smartly):

```javascript
var figures = parseInt(this.expression.attributes.figures, 10);
if(_.isNumber(figures) && !_.isNaN(figues)){
 //...
} else {
  //try to get the variable value
  if(this.state[this.expression.attributes.figures]){
    figures = parseInt(this.state[this.expression.attributes.figures].value, 10);
       if(_.isNumber(figures) && !_.isNaN(figues)){
           //...
       }
   }
}
```

boajer: Check please.

PR #93, opened 2015-02-13
Operator round

boajer:


views/js/scoring/processor/expressions/operators/round.js
```diff
@@ -0,0 +1,82 @@
...
+
+            //if at least one operand is null, then break and return null
+            if(_.some(this.operands, _.isNull) === true){
+                return null;
+            }
+            var value = preProcessor
+                .parseOperands(this.operands).value()[0];
+
+
+            if ( !_.isFinite(value) ) {
```
krampstudio: You should also check for `NaN` and return null

PR #94, opened 2015-02-13
Operator and

boajer:


krampstudio:
Same comments than the OR operator : change min and max and update the tests

PR #99, opened 2015-02-17
operator-durationGTE

boajer:


krampstudio:
You should add duration support in the `typeCaster`.  You can use the library moment-duration (to be included in `tao/views/js/lib/ `) to handle the ISO8601 duration format.

boajer:
Could you clarify, in specification ( http://www.imsglobal.org/question/qtiv2p1/imsqti_infov2p1.html#element10062 ) we have `but represented as a single float that records time in seconds. Durations may have a fractional part. `, so it seems to be not correct?

krampstudio:
Yes, I was confused. Forget my comment.

PR #102, opened 2015-02-18
operator-anyN

boajer:


views/js/scoring/processor/expressions/operators/anyN.js
```diff
@@ -0,0 +1,82 @@
...
+         * @returns {?ProcessingValue} is anyN or null
+         */
+        process : function(){
+
+            var result = {
+                cardinality : 'single',
+                baseType : 'boolean'
+            };
+
+            var min = _.isNumber(this.expression.attributes.min) ? this.expression.attributes.min : 0,
```
krampstudio: As soon as the attributes are `integerOrVariableRef` you should use `this.preProcessor.parseValue(this.expression.attributes.min, 'integerOrVariableRef');` (you may need to merge newrunner into this branch before).
Same for max.

PR #103, opened 2015-02-18
operator-patternMatch

boajer:


krampstudio:
This one is more difficult than it appears. QTI will provide you XML regexp. Please check what is the difference between XML regexp and JS regexp. You'll get them as  strings, so `_.isRegex` will never work.

bugalood:
This link will be a great help http://www.regular-expressions.info/xml.html !

views/js/scoring/processor/expressions/operators/patternMatch.js
```diff
@@ -0,0 +1,83 @@
...
+                baseType : 'boolean'
+            };
+
+            //if at least one operand is null, then break and return null
+            if(_.some(this.operands, _.isNull) === true){
+                return null;
+            }
+
+            //escape $ and ^
+            var pattern = this.expression.attributes.pattern
```
krampstudio: The pattern is of type `stringOrVariableRef` so you should implement it as for `integerOrVariableRef`

PR #113, opened 2015-02-20
operator-multiple

boajer:


views/js/scoring/processor/expressions/operators/multiple.js
```diff
@@ -0,0 +1,88 @@
...
+
+            var acc = [];
+            result.value = preProcessor
+                .parseOperands(filteredOperands)
+                .reduce(function (acc, current) {
+                    acc.push(current);
+                    return acc;
+                }, acc);
+
+            return result;
```
krampstudio: I guess `reduce` isn't useful `value()` should give you the same result

PR #121, opened 2015-02-23
TAO-436: Prevent polygon shape to be resized

hutnikau:
Instead of resizing polyline shape will be transformed by moving the vertex.

### DIFF
--- a/views/js/qtiCreator/widgets/interactions/helpers/shapeEditor.js
+++ b/views/js/qtiCreator/widgets/interactions/helpers/shapeEditor.js
@@ -253,7 +253,6 @@ define([
                  * @private
                  */ 
                 function resized(){
-
                     self.shape.animate(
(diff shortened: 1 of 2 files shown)


clawfire:
It's not working here :/ I got the same behavior than before. When I try to move a vertex the shape just reduce to 0x0. Is there's some commit missing or ... ?

hutnikau:
I checked it and all is works fine. Which browser do you use? Maybe some js files was in cache?
6528a7ad6ed926a010b0122f7485806f

clawfire:
I use Firefox 35.0.1 under osx. I have no cache and use the empty cache command just in case. Don't have this behavior. I'll ask another dev here to test it also ( poke @krampstudio )

clawfire:
And it's works. Thx.

PR #124, opened 2015-02-23
TAO-437: Graphic Gap Match points are too small

hutnikau:
The size of shape handlers always 6 px independently on the shape size.


clawfire:
I think the 6 or 4 px approach is fine. In fact the problem is more about the fact that thoose values are affected by the ratio of `scaleRaphael` and we had to find a way to prevent that. `scaleRaphael` scale everything in raphael for responsive purpose but in this particular case, the handlers have to be the same size everytime.

hutnikau:
Now the size of handlers does not depend on the size of paper.

clawfire:
:+1:

PR #125, opened 2015-02-24
Removed unused custom htaccess

jbout:


ssipasseuth:
Tested and no problem found.

PR #128, opened 2015-02-25
TAO-455 : use CDATA instead of weird  escaping

clawfire:
I test few interaction to see if it's cause any trouble when we enter a response. all looks good. Need more intense testing from our tester.

### DIFF
--- a/test/QtiParsingTest.php
+++ b/test/QtiParsingTest.php
@@ -1,22 +1,22 @@
 <?php
-/**  
+/**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * as published by the Free Software Foundation; under version 2
  * of the License (non-upgradable).
- * 
+ *
  * This program is distributed in the hope that it will be useful,
  * but WITHOUT ANY WARRANTY; without even the implied warranty of
  * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
  * GNU General Public License for more details.
- * 
+ *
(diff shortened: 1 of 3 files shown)


krampstudio:
To be sure it has no impact on the serialization, you should add a test case in `test/QtiParsingTest.php` with a sample that has `CDATA` in the `correctResponse` tag and see if the `responseDeclaration` gives you the correct value.
@ssipasseuth can lead you on adding a test case

clawfire:
@krampstudio done

test/QtiParsingTest.php
```diff
@@ -31,64 +31,93 @@
...
+     * @author Thibault Milan <thibault.milan@vesperiagroup.com>
+     */
+    public function testFileParsingCDATA(){
+        $file = dirname(__FILE__).'/samples/xml/cdata/item.xml';
+
+        $qtiParser = new Parser($file);
+        $qtiParser->validate();
+
+        if(!$qtiParser->isValid()){
+            echo $qtiParser->displayErrors();
```
krampstudio: I would rather use
```php
$this->fail($qtiParser->displayErrors());
```

clawfire: I was just reusing the code just before my test :D But thanks, I'll change that just in a sec

clawfire:
Can't you test if when you save an item, the correct XML is generated **AND** when you leave and come back in authoring, the response is loading in a proper way please ?

clawfire:
Ok I'll adress the ckeditor bug in a separate PR, if this one is ok for you can we merge it ?

PR #129, opened 2015-02-26
operator-mathOperator

boajer:


views/js/scoring/processor/expressions/operators/mathOperator.js
```diff
@@ -0,0 +1,261 @@
...
+        sin: Math.sin,
+        cos: Math.cos,
+        tan: Math.tan,
+        sec: reciprocalFactory(Math.cos),
+        csc: reciprocalFactory(Math.sin),
+        cot: reciprocalFactory(Math.tan),
+        asin: Math.asin,
+        acos: Math.acos,
+        atan: Math.atan,
+        atan2: {
```
krampstudio: Why do you use the property `func` inside ?

boajer: I realized that is not very transparent naming was used there, it is for splitting non standard cases, when function require more than one argument ( like `atan2` ), originally  there are was extra property with number of required arguments

krampstudio: ok I'll just rename it and add a comment

krampstudio:
I've also a test case that fails : 

```
>> Process - mathOperator [exp]
>> Message: The mathOperator value is correct with accuracy 1e-15
>> Actual: 15.154262241479262
>> Expected: 15.154262241479266
```

PR #130, opened 2015-02-26
TAO-351 : Change the styles for the incrementer field 

clawfire:
only in `extendedText` interaction


ssipasseuth:
need to move that interactin specific styles to a dedicated interaction form sass file when the refactoring of the qti creator css is complete

PR #136, opened 2015-02-27
TAO-452: Graphic Gap Match Mime Type should be hidden

hutnikau:


clawfire:
When I click on save the interaction I got a 500 error for invalid item regarding the QTI model. As I said in the card, we must pass this information but we don't want to let the user edit it manually or just see it. 

But it's needed for a valid QTI item.

hutnikau:
I didn't check it with brand new item.
Now it works fine.

PR #137, opened 2015-02-27
operator-inside

boajer:


krampstudio:
The test 'poly outside' is failing

PR #139, opened 2015-03-02
TAO-453: Graphic Gap Match picture scaling

hutnikau:
The Width and Height fields are hiding depending on the `Item width` field value (`Style Sheet Manager` form).

### DIFF
--- a/views/js/qtiCreator/widgets/interactions/graphicGapMatchInteraction/states/Question.js
+++ b/views/js/qtiCreator/widgets/interactions/graphicGapMatchInteraction/states/Question.js
@@ -345,7 +345,8 @@ define([
         var options = widget.options;
         var interaction = widget.element;
         var $form = widget.$form;
-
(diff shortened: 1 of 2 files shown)


views/js/qtiCreator/widgets/interactions/graphicGapMatchInteraction/states/Question.js
```diff
@@ -345,7 +345,8 @@ define([
         var options = widget.options;
         var interaction = widget.element;
         var $form = widget.$form;
-
+        var showWidthInput = $('[name="item-width-prompt"]:checked').val() === "slider";
```
krampstudio: If you want to check whether the item is in responsive mode, you need to check if the interaction contains has the `responsive` class. 
Otherwise your code may break if we change the widget in the panel. 

```
var $container = widget.$original;
var isResponsive = $container.hasClass('responsive');

...
$form.find('.panel-interaction-size').toggle(!isResponsive);
```

hutnikau:
Done.

PR #153, opened 2015-03-10
fixed TAO-473, differences between preview, question and answer mode of ...

dietertaotesting:
...hottext


clawfire:
Code looks good for me and the rendering is what you explain me you want to achieve so :white_check_mark:

PR #160, opened 2015-03-17
TAO-432 get the state and store it instead of response in statevariable ...

antoinerobin:
...to recover the right shuffle


antoinerobin:
https://github.com/oat-sa/extension-tao-item/pull/56

PR #162, opened 2015-03-18
Fix/tao 351 refactor extended text ckeditor

clawfire:


krampstudio:
It seems like line breaks are still allowed while the max word number is exceed.

clawfire:
@krampstudio should be fixed now

PR #163, opened 2015-03-18
Prevent title change on export

jbout:
Don't change the title of the qti item in the itembank after exporting it

### DIFF
--- a/model/Export/QTIPackedItemExporter.php
+++ b/model/Export/QTIPackedItemExporter.php
@@ -105,10 +105,6 @@ class QTIPackedItemExporter extends AbstractQTIItemExporter {
 		
 		if (!is_null($qtiItem)) {
             
-            //tao currently uses the title as the label of imported qti items:
-            $qtiItem->setAttribute('title', $rdfItem->getLabel());
-            $qtiItemService->saveDataItemToRdfItem($qtiItem, $rdfItem);
-            
 		    // -- Prepare data transfer to the imsmanifest.tpl template.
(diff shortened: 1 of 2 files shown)


model/Export/QTIPackedItemExporter.php
```diff
@@ -105,10 +105,6 @@ public function exportManifest() {
 		
 		if (!is_null($qtiItem)) {
             
-            //tao currently uses the title as the label of imported qti items:
-            $qtiItem->setAttribute('title', $rdfItem->getLabel());
-            $qtiItemService->saveDataItemToRdfItem($qtiItem, $rdfItem);
-
```
ssipasseuth: I understand the issue you are trying to solve. However this fix will make an old issue resurface : if a new item has been created in TAO, its initial title is the rdf label (title is required in the standard). However, if the rdf label changes, the item title in the xml is no longer the same. So when we export that item and reimport it back into tao, the rdf resource will not longer have the right value (but the one set upon creation).

ssipasseuth: The real source of the issue is that we want to provide a way to match the rdf item metadata to the one in the qti item xml. The real fix should be to definitely define our policy regarding this matching. What is your opinion on that ?

jbout:
The current behavior is that exporting an item changes the content of the item, not the content of the zip file. I can understand changing the title in the exported XML, but we should not touch the item in the item-bank.

ssipasseuth:
Actually, I am not even sure we should modify the title at all...

PR #164, opened 2015-03-20
jquery version to amd in the newrunner build

krampstudio:
Related to https://github.com/oat-sa/tao-core/pull/195


krampstudio:
Does not require a re bundle

PR #169, opened 2015-03-25
Feature/column layout

ssipasseuth:
This corresponds to TAO-491 : enable inserting multiple row on a column
https://oat-sa.atlassian.net/browse/TAO-491

To test this, you need to activate the multi-column option in the config file : https://github.com/oat-sa/extension-tao-itemqti/blob/feature/column-layout/config/default/qtiCreator.conf.php


config/default/qtiCreator.conf.php
```diff
@@ -0,0 +1,25 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ * 
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ * 
+ * Copyright (c) 2014 (original work) Open Assessment Technologies SA;
```
krampstudio: Are we still in 2014 ? ;)

krampstudio:
Nice work !

krampstudio:
Please note that some comments would be useful on the functions you've added on `creator/main.js`

PR #179, opened 2015-04-03
Refactor/media manager

antoinerobin:


antoinerobin:
depends on https://github.com/oat-sa/tao-core/pull/229

krampstudio:
Some code documentation would have been welcomed to understand the code.

controller/QtiCreator.php
```diff
@@ -143,42 +150,32 @@ public function saveItem(){
         $this->returnJson($returnValue);
     }
 
-    public function getFile(){
+    public function getFile()
```
krampstudio: What happens if the file isn't found ? An exception is thrown ? If yes, IMHO it should be cached and return a 404

PR #181, opened 2015-04-08
TAO-499 added QTI specific add item

jbout:


jbout:
depends on:
https://github.com/oat-sa/tao-core/pull/241
https://github.com/oat-sa/extension-tao-item/pull/64

PR #182, opened 2015-04-08
Removed qti item constant

jbout:


bugalot:
Tested from import to compilation + unit tests (still 1 remaining error as before on devel) -> all good :+1:

PR #188, opened 2015-04-09
TAO-639: TextReader produces a response variable

hutnikau:


bugalot:
Just had a check with the new PR, it's fine with delivery and it solves the issue!
Good work guys!

bugalot:
Because it has to be in master on monday, I validate the PR.

PR #204, opened 2015-04-21
TAO-668	-media element wrapped in AMD-style

boajer:
related to https://github.com/oat-sa/tao-core/pull/281


krampstudio:
It's working great but there is a css issue with the volume button. See http://i.imgur.com/x8OduzT.png 
I guess it's TAO's css that override the player style.

krampstudio:
I've notice also that there a String `video player` at the top of the player. It breaks the layout for example in the resource manager (upload a video and preview it)

krampstudio:
Youtube video does not work anymore in MediaInteraction in Firefox. I've a message in the console saying : 
```
Specified "type" attribute of "video/youtube" is not supported. Load of media resource https://www.youtube.com/watch?v=yLAjKtmT3lk failed.
All candidate resources failed to load. Media load paused.
```

PR #219, opened 2015-04-23
prevent choice changing during elimination

hutnikau:


ssipasseuth:
Is the feature TAO-656 completed ? From what I can see this is but a small part of it.

views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
```diff
@@ -45,7 +45,10 @@ define([
         $container.off('.commonRenderer');
 
         $container.on('click.commonRenderer', '.qti-choice', function(e){
-
+            if ($(this).closest('.qti-itemBody').hasClass('js-choice-eliminating')) {
+                return;
```
ssipasseuth: ok, but isn't it better to run this snippet after e.stopPropagation() (line 54) ?

PR #222, opened 2015-04-28
Do not get the shared libraries of PCI interaction to avoid file not fou...

antoinerobin:
...nd exception


antoinerobin:
depends on : 
https://github.com/oat-sa/extension-tao-mediamanager/pull/18
https://github.com/oat-sa/extension-tao-item/pull/82

ssipasseuth:
This fixes the compilation and export issue.
However, I discovered that import does not work anymore (the problem is already in develop so is not related to this PR).

PR #227, opened 2015-04-29
Capability to disable the tool

hutnikau:
Implementation capability to disable the tool:
Set parameter `disabled` in the `pciCreator.json` (manifest) to true to disable the tool.
Example: 
https://github.com/oat-sa/extension-parcc-student-tools/blob/TAO-655_Student_Tools_placeholders/views/js/picCreator/dev/parccWritingTools/picCreator.json#L21

### DIFF
--- a/views/js/qtiCreator/editor/customElementRegistryFactory.js
+++ b/views/js/qtiCreator/editor/customElementRegistryFactory.js
@@ -45,7 +45,9 @@ define([
             var paths = {};
 
             _(hooks).values().each(function(hook){
-
+                if (hook.manifest && hook.manifest.disabled) {
+                    return;
+                }
                 if(isValidHook(hook)){
 
                     var id = hook.typeIdentifier;


views/js/qtiCreator/editor/customElementRegistryFactory.js
```diff
@@ -45,7 +45,9 @@ define([
             var paths = {};
 
             _(hooks).values().each(function(hook){
-
+                if (hook.manifest && hook.manifest.disabled) {
+                    return;
+                }
                 if(isValidHook(hook)){
```
ssipasseuth: Seems a bit extreme to me, but it does the job for now. I would have suggested to still enable it to be registered and let the hook decide how they should handle disabled custom elements (it may be hidden, it may be grayed out, etc.) With this solution, it is as if they do not exist. But don't worry, we can leave it for now and change it if we need another behaviour.

krampstudio:
- Go the authoring
 - Add the protractor to your item and save
 - Set then the `disabled` prop to `true` for the protractor
 - Open the authoring again and you'll have errors
It's caused by what @ssipasseuth commented. The PIC should be loaded but not visible.

PR #231, opened 2015-05-04
prevent rendering for outcome and response qti classes

ssipasseuth:


ssipasseuth:
this fixes https://oat-sa.atlassian.net/browse/TAO-708

PR #242, opened 2015-05-07
Optimize json structure

ssipasseuth:
This fulfills https://oat-sa.atlassian.net/browse/TAO-646


ssipasseuth:
I added the "performance" results run on 51 sample items in the comments of https://oat-sa.atlassian.net/browse/TAO-646

PR #267, opened 2015-05-21
display the correct path of the file not found

antoinerobin:


bugalot:
Working perfectly! Thanks a lot!

PR #269, opened 2015-05-22
Implementation of list styler into choice interaction

dietertaotesting:
Depends on https://github.com/oat-sa/tao-core/pull/335. /tao-core (develop) must be pretty recent too since it contains the required Javascript


krampstudio:
There's a dependency issue: `ui/listStyler` is missing, even with oat-sa/tao-core#335 merged. Could you reference the PR where it lands ?

krampstudio:
It works as expected except the square in the list : 

Firefox:
firefox

Chrome:
chrome

dietertaotesting:
This square prob. is actually unrelated to this PR. The problem will go away when you resolve the newly created dependency https://github.com/oat-sa/tao-core/pull/356 first.

PR #272, opened 2015-05-27
remove buildAll.php + qtiDefaultRenderer

ssipasseuth:
This concludes the issue https://oat-sa.atlassian.net/browse/TAO-708


krampstudio:
Ho yeah!!!

PR #273, opened 2015-05-27
TAO-782 retrieve xinclude element to export them

antoinerobin:


bugalot:
As agreed, It's OK as a transition. As a final goal, shared stimulus should have a .xml extension instead of .xhtml, because it is valid QTI.

PR #279, opened 2015-06-01
Tao 795 apip authoring api

hutnikau:


views/js/apipCreator/test/api/apipItem.html
```diff
@@ -0,0 +1,75 @@
...
+                            ok(acccessElement.data.localName && acccessElement.data.localName === 'accessElement');
+                            equal(acccessElement.serial, 'accessElement1');
+                            
+                            equal(apipItem.getAccessElementBySerial('wrongSerial'), null);
+                        });
+                        
+                        test("creator.getAccessElementsByInclusionOrder()", function () {
+                            var acccessElement = apipItem.getAccessElementsByInclusionOrder('textOnlyDefaultOrder');
+                            equal(acccessElement.length, 6);
+                            //todo check inclusion order here
```
ssipasseuth: I can see a @todo here : are you planning to do it now or leaving for later ?

views/js/apipCreator/test/api/qtiElement.html
```diff
@@ -0,0 +1,48 @@
...
+        <script  type="text/javascript">
+            require(['/tao/ClientConfig/config'], function () {
+                require(['taoQtiItem/apipCreator/api/qtiElement'], function (QtiElement) {
+                    $.get(
+                        '/taoQtiItem/views/js/apipCreator/test/assets/apip_example_exemplar01.xml'
+                    ).done(function (xml) {
+                        var qtiElement = new QtiElement(xml);
+
+                        test("creator.getAccessElements()", function () {
+                            equal(QtiElement.getAccessElements(), 'itemBody');
```
ssipasseuth: is it actually testing the method getAccessElements() ?

views/js/apipCreator/api/qtiElement.js
```diff
@@ -0,0 +1,67 @@
...
+    }
+
+    /**
+     * Get all the access element(s) related to a qti element
+     * According to the standard, there could more than one access element by qtiElement
+     * 
+     * @returns {Array} array of accessElement
+     */
+    QtiElement.prototype.getAccessElements = function getAccessElements(){
+        return [];
```
ssipasseuth: why are this fn not implemented ?

views/js/apipCreator/api/qtiElement.js
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    /**
+     * Get the unique access element related to a qti element in a specific inclusionOrder
+     * 
+     * @returns {Array} array of accessElement
+     */
+    QtiElement.prototype.getAccessElementByInclusionOrder = function getAccessElementByInclusionOrder(inclusionOrderType){
+        return {};
```
ssipasseuth: why are this fn not implemented ?

views/js/apipCreator/api/qtiElement.js
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    /**
+     * Create an apip accessElement for a qti item body element
+     * 
+     * @returns {Object} the newly create apip access element
+     */
+    QtiElement.prototype.createAccessElement = function createAccessElement(){
+        return {};
```
ssipasseuth: why are this fn not implemented ?

views/js/apipCreator/api/qtiElement.js
```diff
@@ -0,0 +1,67 @@
...
+    };
+
+    /**
+     * Get the order of the qti element, in the order of appearance in the dom of the item body
+     * This is used for inserting a new access element in one of the insertionOrder in a position that reflects the default one
+     * 
+     * @returns {Integer}
+     */
+    QtiElement.prototype.getNativeOrder = function getNativeOrder(){
+        return -1;
```
ssipasseuth: why are this fn not implemented ?

views/js/apipCreator/api/accessElement.js
```diff
@@ -0,0 +1,233 @@
...
+     * @param {object} apipItem apipItem creator api instance 
+     * @see {@link package-tao\taoQtiItem\views\js\apipCreator\api\apipItem.js}
+     * @returns {Object} created XML node
+     */
+    AccessElement.prototype.createXMLNode = function createXMLNode(apipItem) {
+        var accessElementNum = apipItem.xpath('//apip:accessElement').length,
+            accessElementNode = apipItem.createNode(
+                'apip',
+                'accessElement',
+                {"identifier" : ('ae' + accessElementNum)}
```
ssipasseuth: are you sure that this would be unique ? what we have imported an apip item from thrid party-vendors or, if we remove and delete an accessElement [ae0, ae1, ae2], we remove ae1, then we create a new one [ae0, ae2, ae2] 

This reminds me this function : https://github.com/oat-sa/extension-tao-itemqti/blob/master/views/js/qtiItem/core/IdentifiedElement.js#L3-L37
Maybe you may reuse the same idea ? or anything else ?

views/js/apipCreator/api/apipItem.js
```diff
@@ -0,0 +1,212 @@
...
+     * 
+     * @param {String} apipItemXML - the APIP-QTI item XML
+     * @returns {Object}
+     */
+    function ApipItem(apipItemXML) {
+        this.apipDoc = parser.parse(apipItemXML);
+        this.$apipDoc = $(this.apipDoc);
+        this.XMLNS = {
+            'apip': 'http://www.imsglobal.org/xsd/apip/apipv1p0/imsapip_qtiv1p0',
+            'qti': 'http://www.imsglobal.org/xsd/apip/apipv1p0/qtiitem/imsqti_v2p2'
```
ssipasseuth: just a question : would it matter if the qti xml contains namespaced elements such as <m:math>, <xi:include> or <pci:portableCustomInteraction> ?

views/js/apipCreator/api/accessElementInfo/spoken.js
```diff
@@ -0,0 +1,136 @@
...
+            "type" : 'textNode'
+        },
+        "textToSpeechPronunciation.contentLinkIdentifier" : {
+            "type" : 'attribute'
+        },
+        "audioFileInfo.contentLinkIdentifier" : {
+            "type" : 'attribute',
+            "creator" : function (accessElementInfo) {
+                return createAttribute(accessElementInfo, 'audioFileInfo');
+            }
```
ssipasseuth: you may want to create an single function only and pass it as reference for a better readability like that https://gist.github.com/ssipasseuth/e985fc1dc04f1c8191c2

views/js/apipCreator/api/accessElement.js
```diff
@@ -0,0 +1,233 @@
...
+     * 
+     * @param {object} qtiElement QtiElement instance
+     * @returns {Object} ceated <apip:contentLinkInfo> node
+     */
+    AccessElement.prototype.addQtiElement = function addQtiElement(qtiElement) {
+        var that = this,
+            contentLinkInfo = that.apipItem.createNode(
+                'apip',
+                'contentLinkInfo',
+                {"qtiLinkIdentifierRef" : qtiElement.data.getAttribute('id')}
```
ssipasseuth: the attribute id on an qtiElement is required to link it to an accessElement . However it is not mendatory in the qti standard. So should we not check if the attribute id exists for the qtiElement or generate an unique one for the qtiElement here or there https://github.com/oat-sa/extension-tao-itemqti/blob/TAO-795_APIP_authoring_api/views/js/apipCreator/api/qtiElement.js#L79 ?

views/js/apipCreator/api/apipItem.js
hutnikau: Added test case for testing retrieving qti Element with namespace:
https://github.com/oat-sa/extension-tao-itemqti/blob/TAO-795_APIP_authoring_api/views/js/apipCreator/test/api/apipItem.js#L49

https://github.com/oat-sa/extension-tao-itemqti/blob/TAO-795_APIP_authoring_api/views/js/apipCreator/test/assets/apip_example_exemplar01.xml#L37

views/js/apipCreator/api/accessElement.js
hutnikau: What about using Date::getTime() (I mean timestamp)?
Identifiers will be roughly like this:
`ae1433428069128`
`ae1433428069937`

views/js/apipCreator/api/accessElement.js
hutnikau: Fixed:
https://github.com/oat-sa/extension-tao-itemqti/blob/TAO-795_APIP_authoring_api/views/js/apipCreator/api/accessElement.js#L95

test case:
https://github.com/oat-sa/extension-tao-itemqti/blob/TAO-795_APIP_authoring_api/views/js/apipCreator/test/api/accessElement.js#L48

views/js/apipCreator/api/accessElement.js
ssipasseuth: Actually, I find this approach more suitable for serial generation (because it is a system variable).
I would prefer to have a more human readable id for id because those attribute will be exported in the qti xml and id may be used for css selector in styling. So using timestamp for ids is not very practical for human use. That is what I actually meant in the begining (sorry if I has not bean clear enough ...)
Why don't you try to generate an id based one what you are currently doing with the serial, but before assign the id to the node, check if it exists with a global xpath query ?

PR #281, opened 2015-06-04
@font-face in .qti-item now loaded as theme

dietertaotesting:
Depends on https://github.com/oat-sa/tao-core/pull/355


krampstudio:
You need to update the grunt file in order to set up compilation for `views/scss/qti-runner-tao-default.scss`

PR #290, opened 2015-06-10
Tao 797 authoring form

hutnikau:


helpers/Apip.php
```diff
@@ -62,9 +84,13 @@ static public function extractApipAccessibility(DOMDocument $doc)
...
-    {
-        $newNode = $qtiItem->importNode($apipContent->documentElement, true);
-        $qtiItem->documentElement->appendChild($newNode);
+    {        
+        $apipXMLString = $apipContent->saveXML($apipContent->documentElement);
+        
+        $fragment = $qtiItem->createDocumentFragment();
+        $fragment->appendXml($apipXMLString);
+        
+        $qtiItem->documentElement->appendChild($fragment);
```
ssipasseuth: what is the difference between this version (saveXML+createDocumentFragment+appendXml) and the previous one (importNode+appendChild)

hutnikau: Appending apip content using previous approach removes prefixes (`apip:*`) from apip content elements.

ssipasseuth: ok, thanks for the clarification

ssipasseuth:
There are some missing unit tests (for the forms) + styling issues.
That would be done as soon as the libs are proven stable enough.

PR #291, opened 2015-06-10
Fix state erasing after change (deffered call of change event).

jsconan:
The issue was due to deferred event call inside CKEditor engine.
When CKEDITOR.editor.setData() is called, a cascade of event is triggered, and the onchange event is also triggered.

Before my patch, each time the state was set, a reset of the response was performed before the update. The reset forces an erasing of the editor by calling the setData() method, and this method also triggers deferred onchange events, with outdated data in this case. If another call of the setData() method is made just after, the editor is updated twice : first with the new data, then with the reseted data, and the editor is erased.

The solution I found is to avoid to do a reset before updating the response. The reset is only made when the response use a wrong format.

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
@@ -45,7 +45,7 @@ define([
 
         //test when some data is entering in the input field
         //@todo plug the validator + tooltip
-        $element.on('keyup', function(){
+        $element.on('keyup.commonRenderer', function(){
             $element.removeClass('field-error');
             if(!patt.test($element.val())){
                 $element.addClass('field-error');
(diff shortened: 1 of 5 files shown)


krampstudio:
_Sometimes_ (I think when I refresh the page using Shitft-CTRL-R) one of the test is failing: 
> 6 .Extended Text Interaction - XHTML format: renders correctly
> ...
> 11. TypeError: this.getWindow(...).$ is undefined
>Source: 	https://tao.localdomain/tao/views/js/lib/ckeditor/ckeditor.js:75
>Expected 10 assertions, but 11 were run
>Source: 	QUnit.asyncTest@https://tao.localdomain/tao/views/js/lib/qunit/qunit.js:401:3

krampstudio:
There is also another issue: we expect to have a response in XHTML when the format is XHTML (see the [spec](http://www.imsglobal.org/question/qtiv2p1/imsqti_infov2p1.html#element10334)). And we should then restore the state as XHTML.

jsconan:
I made a fix for the XHTML format issue

PR #285, opened 2015-06-08
Tao 819 design the communication api for pic

jsconan:
related to https://github.com/oat-sa/parc-itemrunner/pull/3

As the portable shared libraries are modified, you need to run an update to be able to use the new `listpic` event from the runner provider.

PR #294, opened 2015-06-10
Pci Pic loading

krampstudio:
Change the way a Portable elements are loaded. 
They are now loaded in the same context, the runtime location is either set manually (in authoring and tests) either resolved using the asset manager.

This requires to run the update script in order to register `qtiCusomtInteractionContext` and `qtiInfoControlContext` in the requirejs config.

Related to task TAO-817


manifest.php
```diff
@@ -26,7 +26,7 @@
     'label' => 'QTI item model',
 	'description' => 'TAO QTI item model',
     'license' => 'GPL-2.0',
-    'version' => '2.7.6',
+    'version' => '2.7.7',
```
ssipasseuth: In develop, the current version is already 2.7.7..
This has been updated by the PR : https://github.com/oat-sa/extension-tao-itemqti/pull/285
Should you not update the version to 2.7.8 or more ?

krampstudio:
Oui petit pb de PR concurrentes. Je vais mettre la version à jour.

PR #296, opened 2015-06-11
Feature/import shared stimulus

antoinerobin:


antoinerobin:
https://github.com/oat-sa/extension-tao-itemqti/pull/331

PR #298, opened 2015-06-11
Fix/install portablejs

jbout:


jbout:
should not have been created

PR #312, opened 2015-06-16
TAO-838 if assets are encoded modify src

antoinerobin:
https://github.com/oat-sa/extension-tao-qtiprint/pull/14


krampstudio:
You can't do that. You change the QTI Item model to pass assets. 
As the img url is resolved, you should do it using the strategies.

PR #313, opened 2015-06-16
handle different URLs in stylesheet handler

krampstudio:
- do not bust cache for base64 encoded URLs.
 - do not add the link if there is no href value.
 - minor scope refactoring.


krampstudio:
I've clicked on the "merge" button by inadvertence...  If there are some requested changes I'll add them into another PR.

PR #315, opened 2015-06-17
Ensure asset paths are correct

jbout:
as discussed with @krampstudio


ssipasseuth:
Is this related to a reported issue on Jira ?

jbout:
Not as far as I am aware of,
It is a fix of https://github.com/oat-sa/extension-tao-item/pull/80
and is required for https://github.com/oat-sa/extension-tao-item/pull/105

ssipasseuth:
Ok, thanks.

PR #320, opened 2015-06-19
Prevent the preview ask for save when no changes were made

jsconan:
Prevent the preview asks for save when no changes were made but the loaded item already contains additional styles.


krampstudio:
I've found an issue but it's most likely not due to this PR directly.

PR #323, opened 2015-06-22
count children instead of childNodes to prevent count of textNodes

krampstudio:
A blank line has been added to the qti item template causing the tests to fail in http://build.taotesting.com:8180/job/build-tao-js/177/ 
This PR fixes the way children are counted by ignoring text nodes.


ssipasseuth:
This indeed fixes the assertions on the children however, the test #15 "Provider PIC: Get the list of PIC " is still failing ... 
testfailure

krampstudio:
This test case should be fixed by https://github.com/oat-sa/extension-tao-itemqti/pull/324

PR #324, opened 2015-06-22
Fix/pic state

krampstudio:
Enable to get/set the state of a pic using the standard methods.


views/js/qtiItem/helper/util.js
```diff
@@ -11,6 +33,92 @@ define(['lodash'], function(_){
...
         },
+
+        /**
+         * Generates an id for a Qti element (the generation is different from identifier)
+         * @param {Object} item - the element related item
+         * @param {String} prefix - identifier prefix
+         * @returns {String} the identifier
+         * @throws {TypeError} if there is no item
+         */
+        buildId : function buildSerial(item, prefix){
```
ssipasseuth: You mean buildSerial -> buildId ?

ssipasseuth:
Ok, it looks good to me. Would only need some testing to make sure it works.

jsconan:
The test #16 `Provider PIC: PIC state` never ends under Firefox, but it does under Chrome.
image

There is an error in the console: `TypeError: state.pic is undefined #(test.js:506)`

krampstudio:
ready for a new review;

PR #329, opened 2015-06-25
Tao 910 invalid qti xml

hutnikau:


helpers/Authoring.php
```diff
@@ -123,5 +112,49 @@ public static function addRequiredResources($sourceDirectory, $relativeSourceFil
...
+     * Remove invalid elements and attributes from QTI XML. 
+     * @param string $qti
+     * @return string sanitized XML
+     */
+    public static function sanitizeQtiXml($qti)
+    {
+        $doc = self::loadQtiXml($qti);
+        
+        $xpath = new \DOMXpath($doc);
+        $xpath->registerNamespace('qti', "http://www.imsglobal.org/xsd/imsqti_v2p1");
```
ssipasseuth: what happens if we are using another ns ? like the qtiapip one ?

ssipasseuth: ok, I see that the ns has been hard coded there :  https://github.com/oat-sa/extension-tao-itemqti/blob/master/model/qti/ParserFactory.php#L437

helpers/Authoring.php
```diff
@@ -123,5 +112,48 @@ public static function addRequiredResources($sourceDirectory, $relativeSourceFil
...
+        $dom = new DOMDocument('1.0', 'UTF-8');
+        $dom->formatOutput = true;
+        $dom->preserveWhiteSpace = false;
+        $dom->validateOnParse = false;
+        
+        libxml_use_internal_errors(true);
+        
+        if (!$dom->loadXML($qti)) {
+            $errors = libxml_get_errors();
+            throw new QtiModelException('Wrong QTI item output format');
```
ssipasseuth: Should not we use the $errors to display a more explicit error message in the QtiModelException

helpers/Authoring.php
```diff
@@ -42,40 +42,29 @@ class Authoring
...
-                if(!$parserValidator->isValid()){
-                    common_Logger::w('Invalid QTI output: '.PHP_EOL.' '.$parserValidator->displayErrors());
-                    common_Logger::d(print_r(explode(PHP_EOL, $returnValue), true));
-                    throw new QtiModelException('invalid QTI item XML '.PHP_EOL.' '.$parserValidator->displayErrors());
-                }
-            }
-        }else{
-            $parserValidator = new Parser($qti);
+        //in debug mode, systematically check if the save QTI is standard compliant
+        if(DEBUG_MODE){
```
ssipasseuth: since the validation is already in loadQtiXml, why not simply try and catch the potential exception and throw a user friendly on an upper level ?

jbout:
reverted https://github.com/oat-sa/extension-tao-itemqti/pull/340

PR #331, opened 2015-06-25
add className in getComposingElements to get only some recursive elem…

antoinerobin:
…ents instead of all

### DIFF
--- a/test/QtiModelTest.php
+++ b/test/QtiModelTest.php
@@ -162,6 +162,23 @@ class QtiModelTest extends TaoPhpUnitTestRunner
 
     }
 
+    public function testGetComposingElements(){
+
+        \common_ext_ExtensionsManager::singleton()->getExtensionById('tao');
+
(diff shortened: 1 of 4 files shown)


test/QtiModelTest.php
```diff
@@ -162,6 +162,23 @@ public function testSimpleFeedbackCorrect(){
...
+        $qtiParser = new Parser(dirname(__FILE__).'/samples/xml/qtiv2p1/xinclude/embeded_stimulus.xml');
+        $item = $qtiParser->load();
+
+        $stimulus = $item->getComposingElements('oat\taoQtiItem\model\qti\Xinclude');
+        $this->assertCount(1,$stimulus);
+        $stim = array_shift($stimulus);
+        $this->assertEquals('stimulus.xml',$stim->attr('href'));
+
+        $elements = $item->getComposingElements();
+        \common_Logger::w(print_r(array_keys($elements),true));
```
ssipasseuth: I guess that this warning logging is only  for debugging only right ? so there is no need including that into the final unit test?

PR #343, opened 2015-07-01
Tao 910 sanitize qti xml

hutnikau:
https://github.com/oat-sa/tao-core/pull/405
Must be merged before.


jbout:
Looks good, but as I modified some files I will reassign it to Jerome for independant validation

PR #348, opened 2015-07-06
correct msg plural

dietertaotesting:


views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
```diff
@@ -92,11 +92,12 @@ define([
...
             max = interaction.attr('maxChoices'),
+            msg,
             choiceCount = _.size(interaction.getChoices()),
             minInstructionSet = false;
 
-        //if maxChoice = 1, use the radio gorup behaviour
-        //if maxChoice = 0, inifinite choice possible
+        //if maxChoice = 1, use the radio group behaviour
+        //if maxChoice = 0, infinite choice possible
         if(max > 1 && max < choiceCount){
```
ssipasseuth: I know it has already been like this before, but I think it should be max>=1, otherwise it can never reach the case 'You must select exactly 1 choice'.

dietertaotesting: To be clear, you want the whole condition to be `if(max >= 1 && max < choiceCount){` I suppose? That would make a lot of sense

PR #360, opened 2015-07-10
Fixed errors in console during launch ckeditor

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1039

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
@@ -365,7 +365,7 @@ define([
         var $container = containerHelper.get(interaction);
         var attributes = interaction.getAttributes();
         var responseDeclaration = interaction.getResponseDeclaration();
-        var baseType = responseDeclaration ? responseDeclaration.attr('baseType') : '';
+        var baseType = responseDeclaration.attr('baseType');
         var numericBase = attributes.base || 10;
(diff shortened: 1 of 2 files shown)


views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
```diff
@@ -365,7 +365,7 @@ define([
         var $container = containerHelper.get(interaction);
         var attributes = interaction.getAttributes();
         var responseDeclaration = interaction.getResponseDeclaration();
-        var baseType = responseDeclaration.attr('baseType');
+        var baseType = responseDeclaration ? responseDeclaration.attr('baseType') : '';
```
ssipasseuth: Default baseType to empty string is not a good idea. It may break response format validation during response processing. That said, have you noticed a case when the responseDeclaration is null or undefined ? If it is the case, then finding the source for an empty responseDeclaration is the way to go.

krampstudio: Agree, how `responseDeclaration` can be undefined at that point. May be the work made into https://github.com/oat-sa/extension-tao-itemqti/pull/366 (fix the ready state of interactions) would prevent that case to occur. 
The response cannot be retrieved without declaration.

Alroniks: Thank you @krampstudio I will see your PR, but I work on it now, so have not updated this PR yet.

Alroniks:
Reverted empty baseType. Error with attr should be fixed via https://github.com/oat-sa/extension-tao-itemqti/pull/366, but error with ckeditor during focusing still need to be fixed.
@krampstudio

PR #373, opened 2015-07-22
added user scripts

dietertaotesting:


dietertaotesting:
concerns     TAO-1046

PR #376, opened 2015-07-30
Fix pci interacion import

hutnikau:


model/qti/ImportService.php
```diff
@@ -328,7 +328,7 @@ public function importQTIPACKFile($file, core_kernel_classes_Class $itemClass, $
                         }
                         
                         // replace uri if changed
-                        if ($auxPath != $info['uri']) {
+                        if ($auxPath != ltrim($info['uri'], '/')) {
```
hutnikau: After importing interaction all paths starting with PCI interaction identifier (as described [here](http://forge.taotesting.com/projects/tao/wiki/TAO_PCI_and_PIC#Implementing-pciCreatorjs) we can use relative paths staring with a PCI typeIdentifier) will be prefixed with `/`.
for example:
`<pci:portableCustomInteraction customInteractionTypeIdentifier="adaptiveChoiceInteraction" hook="adaptiveChoiceInteraction/runtime/adaptiveChoiceInteraction.js">`
will be replaced to 
`<pci:portableCustomInteraction customInteractionTypeIdentifier="adaptiveChoiceInteraction" hook="/adaptiveChoiceInteraction/runtime/adaptiveChoiceInteraction.js">`
Because of this replacement all imported PCI interactions are broken.

model/qti/interaction/PortableCustomInteraction.php
```diff
@@ -215,14 +215,17 @@ private function extractPciProperties(DOMElement $propertiesNode, $ns = ''){
...
+    private function serializePciProperties($properties, $ns = '', $name = ''){
         
         $ns = $this->formatPciNs($ns);
-       
-        $returnValue = '<'.$ns.'properties>';
+        if (!empty($name)) {
+            $returnValue = '<'.$ns.'properties key="' . $name . '">';
+        } else {
+            $returnValue = '<'.$ns.'properties>';
+        }
```
hutnikau: During importing PCI interaction `key` attribute of `<properties>` tag is missed.
before importing:
```
<pci:properties key="0">
    <pci:entry key="label">Choice 1</pci:entry>
</pci:properties>
```
after importing:
```
<pci:properties>
    <pci:entry key="label">Choice 1</pci:entry>
</pci:properties>
```

model/qti/interaction/PortableCustomInteraction.php
```diff
@@ -215,14 +215,17 @@ private function extractPciProperties(DOMElement $propertiesNode, $ns = ''){
...
+    private function serializePciProperties($properties, $ns = '', $name = null){
         
         $ns = $this->formatPciNs($ns);
-       
-        $returnValue = '<'.$ns.'properties>';
+        if ($name !== null) {
+            $returnValue = '<'.$ns.'properties key="' . $name . '">';
+        } else {
+            $returnValue = '<'.$ns.'properties>';
+        }
```
hutnikau: During importing PCI interaction key attribute of <properties> tag is missed.
before importing:
```
<pci:properties key="0">
    <pci:entry key="label">Choice 1</pci:entry>
</pci:properties>
```
after importing:
```
<pci:properties>
    <pci:entry key="label">Choice 1</pci:entry>
</pci:properties>
```

bugalot:
Hello there !

Thanks @hutnikau for this that's great job! Using this branch, I could import back the scaffolding_images.zip from JIRA TAO-1013 archived item and make it fully working in preview mode, including scoring, choice masking etc... There is just a little glitch remaining which I will describe right now in JIRA TAO-1013.

@ssipasseuth we will have to make sure this does not break anything with standard QTI Items assets (img, css, ...) with respect with the "verbatim" TAO and/or the MediaManager. Could you please advise?

Cheers,
Jérôme

hutnikau:
I've found one more issue that occurs during import:
before import:
```
<pci:entry key="label">&lt;div&gt;
&lt;div&gt;Choice 1&lt;img src="091210e755b736db3ff5a.jpg" alt="nature-q-c-800-600-10.jpg" width="100%"  /&gt;&lt;/div&gt;
&lt;/div&gt;
</pci:entry>
```
after import:
```
<pci:entry key="label">
    <div>
    <div>Choice 1<img src="091210e755b736db3ff5a.jpg" alt="nature-q-c-800-600-10.jpg" width="100%"/></div>
    </div>
</pci:entry>
```
After import all html entities in the properties are decoded (but it have to be encoded). Because of that when new imported item opened first time images is not visible.

bugalot:
Maybe a simple fix to clean-up PCI entry elements could be

```php
// $xpath instanceof DOMXPath.
$xpath->registerNamespace('pci', 'http://blabla...');
$entries = $xpath->query('//pci:entry');
$ownerDoc = $xpath->document;

foreach ($entries as $entry) {
    $strContent = '';
    foreach ($entry->childNodes as $child) {
        $strContent .= $ownerDoc->saveXML($child);
    }

    $entry->parentNode->replaceChild(
        $ownerDoc->createTextNode($strContent);
        $entry
    );
}
```

Not tested :relaxed: !

model/qti/ImportService.php
ssipasseuth: looks a good idea but I am just curious : by any chance; would you know when and where the / has been added ?

model/qti/interaction/PortableCustomInteraction.php
ssipasseuth: ok, nice catch !

model/qti/ImportService.php
hutnikau: Yes, i know.
This slash added here:
https://github.com/oat-sa/extension-tao-item/blob/develop/models/classes/media/LocalItemSource.php#L162

ssipasseuth:
I can see that there are numerous cases which shows that the unit tests are not sufficient. Could you help by adding additional test samples and unit test them please ?

ssipasseuth:
@bugalot I believe that more samples in the unit test would help figuring out if that might break the standard item or not. But in any case, I don't think that it will ever make sense to have / before any url in the xml since we don't know what the / would be refering to anyway.

hutnikau:
I added unit test to check importing PCI interaction.

hutnikau:
I've replaced  `htmlentities()` to `htmlspecialchars()`.

ssipasseuth:
It appears that ENT_XML1 is only available from php 5.4 but we need also to support php 5.3. So could you please another solution.
On top of that, I would also suggest you add an unit test to check this html entities issue.

bugalot:
Guys, again, why not using DOMDocument::saveXML to serialize a DOMText. It's built-in ;) !

hutnikau:
I agree that `DOMDocument::saveXML()` can be useful there but there is recursive function which returns only fragment of document. In any case I will try to use it.

hutnikau:
Done...

model/qti/ImportService.php
ssipasseuth: Ok, ok I see, thanks !

ssipasseuth:
Ok, let me check again

PR #377, opened 2015-07-31
Fix/missing shared file

antoinerobin:


antoinerobin:
depends on https://github.com/oat-sa/tao-core/pull/454

PR #396, opened 2015-08-12
Moved loader to tao/ui

dietertaotesting:
depends on #481


dietertaotesting:
depends on  https://github.com/oat-sa/tao-core/pull/481

PR #407, opened 2015-08-20
resolve load/setData race condition

krampstudio:
extended text interaction with xhtml format and word/char count constraint to be tested (along with the unit test)


views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
```diff
@@ -306,12 +312,7 @@ define([
         return ret;
     };
 
-
-    /**
-     * Limit the input (words, chars, etc.)
-     * @param {Object} interaction - the extended text interaction model
-     */
-    var limitUserInput = function limitUserInput(interaction){
+    var inputLimiter = function userInputLimier(interaction){
```
jsconan: Doc annotations were removed, is it wanted ?

PR #408, opened 2015-08-20
Tao 1272/svgz aac support

antoinerobin:


views/js/qtiCreator/widgets/static/img/states/Active.js
```diff
@@ -200,7 +200,7 @@ define([
                 params : {
                     uri : options.uri,
                     lang : options.lang,
-                    filters : 'image/jpeg,image/png,image/gif'
+                    filters : 'image/jpeg,image/png,image/gif,image/svg+xml,application/x-gzip'
```
krampstudio: I have a small issue with this mime type. Does it means we can also upload any .gz file?

antoinerobin:
depends on : https://github.com/oat-sa/tao-core/pull/504

PR #409, opened 2015-08-21
Introduced programmatic fallback for unsupported in FF@Macos focus event

boajer:
depends on https://github.com/oat-sa/extension-tao-parcc/pull/17

### DIFF
--- a/config/default/userScripts.conf.php
+++ b/config/default/userScripts.conf.php
@@ -22,4 +22,8 @@
 /**
  * Default user scripts, should be empty
  */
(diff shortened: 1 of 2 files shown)


krampstudio:
I've a problem with this implementation:
 - they won't benefit the fix in their test driver (they don't have user scripts)
 - if tomorrow the issue is fixed by the browsers, you'll still fallback to js (browser based detection vs feature detection)
 - it rely on a class that is defined on a customer theme...

Let us think about that before

PR #417, opened 2015-08-27
Added tabindex attributes to several interactions for proper work of …

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1034


Alroniks:
Rebased, because was conflict with this merged PR https://github.com/oat-sa/extension-tao-itemqti/pull/418

krampstudio:
Would'nt it be better to have `tabindex="0"` so we can keep tabbing in the order of appearance.

Alroniks:
Yes, I agree @krampstudio Updated PR

krampstudio:
The inlineChoice interaction doesn't focus correctly : 
 - you have to tab 2 times
 - the style is not applied (the shady border)

Alroniks:
@krampstudio which browser do you use? I've checked in Chrome and Firefox on Mac.

krampstudio:
I'm able to reproduce with both.

Alroniks:
Reverted 1 to 0 for tabindex.

Alroniks:
Close for https://github.com/oat-sa/extension-tao-itemqti/pull/424

PR #425, opened 2015-09-04
Fixed structure for qti.xml in 'Planets and Moons'

boajer:
Here was removed extra wrapping for interaction, that prevents normal deleting it from item


ssipasseuth:
The issue is not with the qti item. The item is perfectly valid qti item.
The problem lies with the iteam creator and it is indeed a bug. Please see my suggested fix https://github.com/oat-sa/extension-tao-itemqti/pull/426

PR #429, opened 2015-09-10
return viewport height if no footer

hutnikau:


ssipasseuth:
As discussed yesterday, there must always be a footer. The bug is therefore not in the qti creator but on the missing template (that was my fault, sorry! ) I fix it there https://github.com/oat-sa/tao-core/pull/533
So no need for this actually.

PR #435, opened 2015-09-22
Add validatior to spoken form

hutnikau:
related:
https://github.com/oat-sa/tao-core/pull/554 (merged)
https://github.com/oat-sa/tao-core/pull/549


ssipasseuth:
I have an issue when I try to modify and form element : input,  textarea, dropdown for the audio files.
When I try to set focus on them, I lose focus and it seems that the form blinks because of reinitialization ?

ssipasseuth:
Another issue is related to the user experience.
To me, the editor should be wysiwyg: so when I see the form prefilled (which is nice !) I expect my access element to be "almost" created. So unless I click on cancel, I expect the access element to be created.
IN such a situation, the success feedback that shows that the access element has been created looks strange to me.
What do you think ?

hutnikau:
You can click on cancel, delete or on another qti element. In all this situations the access element will not be created/saved. The only way to save the access element is click on Done button. So, i think that feedback message should be shown to make user know that form not just closed, but saved and closed.

ssipasseuth:
mmm, I understand your point.
I merge the TAO-779-apip-authoring tool in it. The behaviour is not good enough. But I will find a solution.
For now, I have noted a few smaller improvements that I can suggest.
1. the error tooltip z-index is too high , it goes above the resource manager (when you select an audio file)
2. I believe we should remove the error feedback when user is trying to correct the error (let's be indulgent, he is trying to fix his error but the red message only disappears on blur)

hutnikau:
PR has been updated. I changed event type of of validation (to `keyup`).
Bug with z-index still not fixed. I will fix it when pull requests with qtip plugin will be merged:
https://github.com/oat-sa/tao-core/pull/630
I just will need to change type of error highlighter.

ssipasseuth:
APIP project has been put aside for a while, let's close this PR for now not to clutter our list of PRs.

PR #439, opened 2015-09-25
Define in real percents width of center button for avoid auto calcula…

Alroniks:
…tion in some cases

Related to https://oat-sa.atlassian.net/browse/TAO-1497


dietertaotesting:
I won't merge this in the main product but add to the concerned theme

PR #442, opened 2015-09-29
Tao 850 apip ordering

ssipasseuth:


helpers/Apip.php
```diff
@@ -65,7 +65,7 @@ static public function extractQtiModel(DOMDocument $doc)
     {
         $accessibilityNode = $doc->getElementsByTagName('apipAccessibility');
         if ($accessibilityNode->length) {
-            $accessibilityNode[0]->parentNode->removeChild($accessibilityNode[0]);
+            $accessibilityNode->item(0)->parentNode->removeChild($accessibilityNode->item(0));
```
ssipasseuth: I also found that issue that causes an error when I save from my desktop PC

views/js/apipCreator/editor/inclusionOrderListing.js
```diff
@@ -0,0 +1,81 @@
...
+    'use strict';
+    
+    var _ns = '.inclusion-order-listing';
+    
+    function render($container, elements){
+        
+        var i = 0;
+        var _elements = _.map(elements, function(element){
+            i++;//starting from 0
+            return {
```
hutnikau: I suggest use the second callback function parameter as index instead of incrementing `i` variable:
```
_.map(elements, function(element, i){
  return {
    id : element.id,
    qti : element.qti,
    content : element.content,
    order : (i + 1) //starting from 1
  };
});
```

views/js/apipCreator/editor/inclusionOrderListing.js
```diff
@@ -0,0 +1,81 @@
...
+            change : function(e, ui){
+                
+                var $helper = $(ui.helper);
+                var i = 1;
+                var qtiOrder = [];
+                var aeOrder = [];
+                
+                $sortable.children('.order-element').not('.ui-sortable-helper').each(function(e){
+                    
+                    var $li = $(this);
```
hutnikau: The same thing here. I suggest use first callback function parameter as index instead of incrementing `i` variable:
```
.each(function(i){
    ...
    $li.data('order', i);
    $li.find('.order').html(i);
    ...
});
```

views/js/apipCreator/editor/qtiElementSelector.js
```diff
@@ -217,7 +220,11 @@ define([
                 var qtiElements = ae.getQtiElements();
                 _.each(qtiElements, function (qtiElement){
                     //set feature css class to qti element
-                    $container.find('.element[data-serial="' + qtiElement.serial + '"]').addClass('apip-feature-info-' + aeInfoType);
+                    $container.find('.element[data-serial="' + qtiElement.serial + '"]')
+                        .addClass('apip-feature-info-' + aeInfoType)
+                        .find('.order').addClass('fill').html(i);
```
hutnikau: After i select qti element which contains few other elements then all nested elements are numerated.

apip_numerated_elements


I guess only direct children should be numerated here:
```
$container.find('.element[data-serial="' + qtiElement.serial + '"]')
    .addClass('apip-feature-info-' + aeInfoType)
    .children('.information').find('.order').addClass('fill').html(i);
```

ssipasseuth:
Ok, ready for review.
Not necessary for now, but in the future, I would rather imagine the ordering happening in the left of the authoring tool instead of the right, what do you think about that ?

hutnikau:
Thank you. All works fine!

PR #446, opened 2015-10-02
Tao 1502 layout options

dietertaotesting:


dietertaotesting:
Actual usage in https://github.com/oat-sa/extension-tao-mp/pull/46

PR #452, opened 2015-10-12
Discard textReaderInteraction from sending response

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-829


ssipasseuth:
Ok, works for depp as a hot fix.

PR #458, opened 2015-10-14
Trial to have correct messages when items fail to import.

bugalot:
This PR inspects the parser's source to determines which is the correct XSD to be used for validation at QTI Item Package import time.

Antoine, I would like you to check

1. When importing an invalid QTI Package (select QTI in import form), the error message is correctly displayed and appropriate.
2. When importing an invalid APIP Package (select APIP in import form), the error message is correctly displayed and appropriate.
3. When importing a valid QTI Package, there is no error message and everything is there.
4. When importing a valid APIP Package, there is no error message and everything is there.

I send you some samples by email.

### DIFF
--- a/model/qti/Parser.php
+++ b/model/qti/Parser.php
@@ -55,11 +55,36 @@ class Parser extends tao_models_classes_Parser
     public function validate($schema = ''){
         
         if (empty($schema)) {
-            $validSchema = $this->validateMultiple(array(
-                __DIR__.'/data/qtiv2p1/imsqti_v2p1.xsd',
-                __DIR__.'/data/qtiv2p0/imsqti_v2p0.xsd',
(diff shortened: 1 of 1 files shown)


model/qti/Parser.php
```diff
@@ -55,11 +55,36 @@ class Parser extends tao_models_classes_Parser
...
-            ));
+            
+            // Let's detect NS in use...
+            $dom = new DOMDocument('1.0', 'UTF-8');
+            $dom->loadXML($this->source);
+            
+            // default is QTI 2.1.
+            $schemas = array(
+                __DIR__.'/data/qtiv2p1/imsqti_v2p1.xsd'
+            );
```
antoinerobin: You could put it in the default of the switch to be consistent.

PR #460, opened 2015-10-19
made the alt of an image optional in the authoring

ssipasseuth:
It makes the ALT attribute optional in the authoring tool.
However it still rendered an empty string in qti xml because ALT is required by the standard.


bugalot:
Did the whole loop:

1. Authoring with empty alt = OK
2. Preview = OK
3. Compile and take test = OK
4. Inspect exported item = OK

PR #471, opened 2015-10-29
Added call waitForMedia before call resize to be sure that all inner …

Alroniks:
…elements loaded

Related to https://oat-sa.atlassian.net/browse/TAO-1691


Alroniks:
Here not implemented support for video in waitForMedia, it should be done here or in another PR.

PR #472, opened 2015-10-30
Refactoring adaptive sizing of interactions in edit and preview mode

Alroniks:
Related to https://github.com/oat-sa/extension-tao-itemqti/pull/471


Alroniks:
Updated after discussion with Dieter.

PR #478, opened 2015-11-06
modify xmlns to ns to make xpath works

antoinerobin:


krampstudio:
I've merged it on develop

PR #480, opened 2015-11-11
Notice prevented when no items selected during class export

boajer:


antoinerobin:
still have the error ```tao/actions/class.RdfController.php at line 125: Missing or invalid parameter uri```when exporting item's class

boajer:
Oh, actually that is error is fixed by https://github.com/oat-sa/tao-core/pull/628. 
And current PR is more related to bug that happen on the same page, but if user trying to export class after clicking on uncheck all link. Just fix it passing by.

antoinerobin:
Ok but doing this you now have in logger a ```Missing filename for export using oat\taoQtiItem\model\Export\QtiPackageExportHandler```because it goes to the else, you should differentiate the two errors.

boajer:
Good point, i've fixed that embarrassing mistake

PR #481, opened 2015-11-12
Removed ref to qti.scss from build

dietertaotesting:


krampstudio:
It seems he file `views/css/qti.css` is still used in : 
 - `views/templates/QtiCreator/index.tpl`
 - in most of the tests 

We need to align the source, the target and the build

PR #483, opened 2015-11-17
Added wrapped string with mathml expression by parseHTML for save pro…

Alroniks:
…cessing with broken xml/html

Related to https://oat-sa.atlassian.net/browse/TAO-1735

### DIFF
--- a/views/js/qtiItem/core/Math.js
+++ b/views/js/qtiItem/core/Math.js
@@ -44,7 +44,7 @@ define([
         var hasContent = false;
 
         if(mathStr && mathStr.trim()){
-            var $math = $(mathStr);
+            var $math = $($.parseHTML(mathStr));
             hasContent = !!$math.text();
         }
 


ssipasseuth:
Hello, sorry, can you tell me the exact steps to follow in order to reproduce the initial issue reported in TAO-1735 ? 
Actually, I cannot, that is why I am having issue making the connection with your suggested solution to this that issue.
Thanks.

Alroniks:
Yes, no problem 
https://www.youtube.com/watch?v=RZnim9ONgfw

Alroniks:
Problem, that when user start type tags, _isEmptyMathML function try get text from input field and create jquery object from it. But '<' not valid for jquery and I added parseHTML() for safe wrap text from user.

krampstudio:
Be careful with `parseHTML` as it uses DOM elements. So if you parse MathML that is not supported by the browser it may break. Just double check on different browsers.

ssipasseuth:
Thanks for all your clarification, I could not reproduce it because I was too lazy to type all the mathML xml in so I just copied and pasted the mathML sample into the authoring tool.
@Alroniks , can you confirm that this solution is fully functionnal on all browsers as @krampstudio underlined ?

Alroniks:
@ssipasseuth Yes, it's bug appears only when typing by hands. I've checked in Chrome, Firefox and Safary, need check in IE/Edge, but my VM with windows is broken, I need time to restore it.

Alroniks:
@ssipasseuth I've cheked in IE 11 in windows 7 - all works fine. Not have windows 10 and Edge for check.

Alroniks:
Also cheked in Edge in win10 and all works fine. @ssipasseuth

ssipasseuth:
@Alroniks thanks for your extensive test reports. I have tested also on win10(win7) and chromium and latest ff on ubuntu. Looks ok.

PR #484, opened 2015-11-18
removed max height/width from associate interaction

dietertaotesting:


dietertaotesting:
Create an associate interaction with an element higher than 150px. If you preview the item in `develop` you will note that the target containers are not high enough, whereas in this branch they are.

PR #490, opened 2015-11-23
Removed reserved classnames String,Integer,Float,Boolean

jbout:


Alroniks:
+1 for merge this.

boajer:
+1 I'm look forward to these changes

bugalot:
I agree too. However, I'm not comfortable and familiar enough with this version of the QTI model to make sure everything is fine.

@ssipasseuth What could be the impact on such a change on the QTI Item php model ?

PR #494, opened 2015-12-01
Compile scripts

Alroniks:


llecaque:
recompilation done in https://github.com/oat-sa/extension-tao-itemqti/pull/496

PR #498, opened 2015-12-09
discarding the choice now re-adds .qti-choice

dietertaotesting:


ssipasseuth:
If it is a hotfix, should the client bundle need to be rebuilt ?

dietertaotesting:
I have added the bundle to https://github.com/oat-sa/extension-tao-itemqti/pull/501/files also

PR #500, opened 2015-12-09
Removing comma from number in text entry interaction for valid parsing

Alroniks:
Related to https://oat-sa.atlassian.net/browse/TAO-1726
And also description in comments here https://oat-sa.atlassian.net/browse/TAO-1723

parseFloat works only with dot in numbers, independently from user locale, so I think remove comma will be enough and better then include big library for this small operation.

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js
@@ -29,9 +29,10 @@ define([
     'taoQtiItem/qtiCommonRenderer/helpers/container',
     'taoQtiItem/qtiCommonRenderer/helpers/instructions/instructionManager',
     'taoQtiItem/qtiCommonRenderer/helpers/PciResponse',
+    'util/locale',
     'polyfill/placeholders',
     'tooltipster'
-], function($, _, __, tpl, containerHelper, instructionMgr, pciResponse){
+], function($, _, __, tpl, containerHelper, instructionMgr, pciResponse, locale){
     'use strict';
 
     /**
@@ -146,11 +147,11 @@ define([
             //invalid response or response equals to the placeholder text are considered empty
             value = '';
         }else{
-            if(baseType === 'integer'){
-                value = parseInt($el.val(), numericBase);
-            }else if(baseType === 'float'){
-                value = parseFloat($el.val());
-            }else if(baseType === 'string'){
+            if (baseType === 'integer') {
+                value = locale.parseInt($el.val(), numericBase);
+            } else if (baseType === 'float') {
+                value = locale.parseFloat($el.val());
+            } else if (baseType === 'string') {
                 value = $el.val();
             }
         }
--- a/views/js/test/runner/interactions/textentry/test.js
+++ b/views/js/test/runner/interactions/textentry/test.js
@@ -283,5 +283,6 @@ define([
             .init()
             .render(container);
     });
+
 });
 


views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js
```diff
@@ -149,7 +149,7 @@ define([
             if(baseType === 'integer'){
                 value = parseInt($el.val(), numericBase);
             }else if(baseType === 'float'){
-                value = parseFloat($el.val());
+                value = parseFloat($el.val().replace(/,/g, ''));
```
krampstudio: I'd prefer that you remove it before the parse : 
```js
var val = "5, 000, 000.15;
parseFloat(val.replace(/,/g, '')) === 5000000.15;
```

krampstudio:
Please add a unit test in https://github.com/oat-sa/extension-tao-itemqti/tree/develop/views/js/test/qtiCommonRenderer/interactions 
The test doesn't need to cover everything (you can use another one as example), but at least your case.

Alroniks:
Rebased branch from current develop and added unit tests. @krampstudio Please, check again

ssipasseuth:
Hi, I checked your suggested fix but after discussing it with Tom, we still prefer to have a way to configure the decimal separator on the plateform level. Indeed the current implementation will break for half of the world. https://en.wikipedia.org/wiki/Decimal_mark#Hindu_numeral_system
In that configuration, the default separator should be a dot, but it should be configurable to comma for e.g. European countries.

ssipasseuth:
Technical advice
1. I would suggest you create a generic localization helper in tao/views/js/util/local.js
This way we can extend the helper with other features
2. Start by providing a first method getDecimalSeparator() that will return whatever the configuration on the server will be
3. To make this tao/util/local module configurable server-side please use the oat\tao\model\ClientLibConfigRegistry (see an exemple of usage here https://github.com/oat-sa/extension-tao-act/blob/develop/scripts/install/registerQtiItemScripts.php)
4. Don't forget update scripts
5. Don't forget to unit test it thoroughly

That will make it a more complete solution and we are aware that will take you extra time.

Thanks!

Alroniks:
Thank you @ssipasseuth for clarification.

ssipasseuth:
You are welcome :)

ssipasseuth:
This is a bug fix for project PARCC. And I just checked with @jbout, the fix needs to be done to the master branch for extensions that are not parcc related. So could you please create a branch from master, cherry pick your changes and create a PR to master. (also please do not increment the manifest extension after a fix for tracing purpose, even if there is no update script to execute)

ssipasseuth:
Replaced with TAO-1726_localised_number_parser

jbout:
Actually please DO increment the manifest extension, as we need to create a
tag for this bugfix and ensure the update script takes this into account

On 20 January 2016 at 10:47, sam <notifications@github.com> wrote:

> Replaced with TAO-1726_localised_number_parser
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/oat-sa/extension-tao-itemqti/pull/500#issuecomment-173147519>
> .
>



-- 





              Bout Joel
*
Tao Architect*



                    *TAO S.A.*

                    9 avenue des Hauts-Fourneaux,

                    L-4362 Esch-sur-Alzette in Belval



                    tél: +352-545580-592

                    e-mail: joel.bout@taotesting.com


                    Visit http://www.taotesting.com

PR #501, opened 2015-12-09
discarding the choice now re-adds .qti-choice

jbout:
Please reassign as required


ssipasseuth:
This does not look good, can you please cherry pick the one commit to that fix ?

jbout:
This was a cherry pick, no idea what happened here

PR #503, opened 2015-12-15
TAO-1759, wrapped img in promise

dietertaotesting:


hutnikau:
I still can reproduce this bug by limiting network speed in the devtools. I guess this occurs not due to images (not only due to images?) but due to choices. I have noticed that at the beginning choices height is thinner than after loading complete:
choices1
and in few seconds the choices became thicker:
choices2
May be some classes adds to the choices or css file loads and changes it's height.

PR #504, opened 2015-12-16
do not render empty value

hutnikau:
If correct value has not been selected then value element will look like this:
```
<value fieldIdentifier="actualResponse" baseType="integer"><![CDATA[]]></value>
```
to prevent rendering element with empty value I added this condition.


ssipasseuth:
If this need to go into mpdeploy, you need to make a PR to the mpdeploy branch. Please confirm if this is needed for mpdeploy or not. Thanks

ssipasseuth:
Don't you need also to update the php template too ?

ssipasseuth:
If this need to be deployed, please consider rebuilding the client bundle

PR #505, opened 2015-12-17
do not render empty value

hutnikau:
related to:
https://github.com/oat-sa/extension-tao-itemqti/pull/504

### DIFF
--- a/views/js/qtiXmlRenderer/tpl/responseDeclaration.tpl
+++ b/views/js/qtiXmlRenderer/tpl/responseDeclaration.tpl
@@ -12,7 +12,9 @@
     <correctResponse>
         {{~#each correctResponse}}
         {{~#if ../isRecord}}
+        {{~#if value}}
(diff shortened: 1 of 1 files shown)


ssipasseuth:
why there is no need to update the template on the server as for https://github.com/oat-sa/extension-tao-itemqti/pull/504 ?

hutnikau:
mpdeploy branch has different template on the server side.

ssipasseuth:
not sure I understand, do you mean https://github.com/oat-sa/extension-tao-itemqti/blob/TAO-1914_record_response_type_mpdeploy/model/qti/templates/qti.responseDeclaration.tpl.php ? It is identical in both develop and mpdeploy branches.

hutnikau:
No, it is not identical:
https://github.com/oat-sa/extension-tao-itemqti/blob/mpdeploy/model/qti/templates/qti.responseDeclaration.tpl.php
https://github.com/oat-sa/extension-tao-itemqti/blob/develop/model/qti/templates/qti.responseDeclaration.tpl.php

ssipasseuth:
So is it alright that mpdeploy won't support record type correct response declaration for now ?

PR #510, opened 2015-12-23
empty elements are now closed correctly

dietertaotesting:


model/qti/container/Container.php
```diff
@@ -188,6 +190,29 @@ public function checkIntegrity($body, &$missingElements = null){
     }
 
     /**
+     * Converts <foo /> to <foo></foo> unless foo is a proper void element such as img etc.
+     *
+     * @param $html
+     * @return mixed
+     */
+    public function fixNonvoidTags($html) {
+        return preg_replace_callback('~(<([\w]+)[^>]*?)(\s*/>)~', function ($matches) {
```
bugalot: would use flag 'u' to no break unicode.

model/qti/container/Container.php
```diff
@@ -188,6 +190,29 @@ public function checkIntegrity($body, &$missingElements = null){
...
+                // do nothing
+                return $matches[0];
+            }
+            // regular void elements
+            if(in_array($matches[2], array('area','base','br','col','embed','hr','img','input','keygen','link','meta','param','source','track','wbr'))) {
+                // do nothing
+                return $matches[0];
+            }
+            // correctly closed element
+            return trim(substr($matches[0], 0, -2)) . '></' . $matches[2] . '>';
```
bugalot: PHP's substr is not binary-safe. Please use mb_substr with 'UTF-8' encoding to make sure we do not break funny characters.

model/qti/container/Container.php
```diff
@@ -150,6 +150,8 @@ public function getBody(){
      * modify the content of the body
      * 
      * @param string $body
+     * @param bool $integrityCheck
+     * @return bool
      */
     public function edit($body, $integrityCheck = false){
```
bugalot: Could we just add a call to the common logger to identify the content is being edited programmaticaly?

bugalot:
I really like you @dietertaotesting, however, I cannot merge this because it acts not only on the HTML subset of QTI but on the whole set of elements. The preg_replace logic is the good one, so you can reuse it in a more appropriate place. Maybe, as we discussed, in the Packer.

Warm regards,
Jérôme

PR #524, opened 2016-01-14
add the possibility to configure the item compiler

antoinerobin:
add a new compiler to compile items in Json format

By modifying config/taQtiItem/compilerClass.conf.php you can choose you qti item compiler class


scripts/update/Updater.php
```diff
@@ -185,7 +185,15 @@ public function update($initialVersion){
             $currentVersion = '2.12.0';
         }
 
-        return $currentVersion;
+        $this->setVersion($currentVersion);
+
+        if($this->isVersion('2.12.0')) {
+            $itemQtiExt = \common_ext_ExtensionsManager::singleton()->getExtensionById('taoQtiItem');
+            $compilerClassConfig = 'oat\taoQtiItem\models\QtiItemCompiler';
```
jsconan: Wrong namespace: this is `model`, not `models`

PR #545, opened 2016-01-28
Fix/filtered item data

antoinerobin:


krampstudio:
to be tested with modal feedbacks

PR #552, opened 2016-01-31
prevent error whithout a type

krampstudio:
Prevent the interaction to break the runtime when the attr is missing.


dietertaotesting:
https://github.com/oat-sa/extension-tao-itemqti/pull/543 already addresses this problem. Closing this PR

PR #570, opened 2016-02-24
Feature/extension installed event

antoinerobin:


antoinerobin:
depends on : 
* https://github.com/oat-sa/extension-tao-item/pull/181
* https://github.com/oat-sa/tao-core/pull/745

PR #576, opened 2016-03-02
fixed wrong response submitted for multiple slider interactions

ssipasseuth:


krampstudio:
Aren't we suppose to use a container and a serial instead of an `id` in the DOM ? Even if you fix the issue I think this interaction should be refactored (and have a dedicated test).

PR #578, opened 2016-03-02
Feature/tao 2338 french translation

siwane:


llecaque:
need to be regenerate

PR #583, opened 2016-03-08
Fix/tao 2310 mathjax bug

no-chris:
How to test:
- add a mathjax latex element
- add the following string: 
- \frac12
- insert a < before the a:
- \fr<ac12
- with this fix, editor shouldn't become unresponsive anymore and we can still use Mathjax (edit current expression, create new ones...)


no-chris:
closing because I chose the wrong target branch, will open a new one

PR #584, opened 2016-03-09
Fix/tao 2310 mathjax bug

no-chris:
How to test:
- add a mathjax latex element
- add the following string:
- \frac12
- insert a < before the a:
- \fr<ac12
- with this fix, editor shouldn't become unresponsive anymore and we can still use Mathjax (edit current expression, create new ones...)

### DIFF
--- a/views/js/qtiCreator/editor/MathEditor.js
+++ b/views/js/qtiCreator/editor/MathEditor.js
@@ -96,11 +96,11 @@ define(['lodash', 'jquery', 'mathJax'], function(_, $, MathJax){
     MathEditor.prototype.renderFromTex = function(){
 
         var args = _processArguments(this, arguments);
+        var _this = this;
+        var jaxQueue = MathJax.Hub.queue;
 
-        if(typeof(MathJax) !== 'undefined'){
+        if(typeof (MathJax) !== 'undefined'){
 
-            var _this = this;
-            var jaxQueue = MathJax.Hub.queue;
             if(this.display === 'block'){
                 _this.$buffer.text('\\[\\displaystyle{' + _this.tex + '}\\]');
             }else{
@@ -110,14 +110,14 @@ define(['lodash', 'jquery', 'mathJax'], function(_, $, MathJax){
             //render preview:
             jaxQueue.Push(
                 //programmatically typeset the buffer
-                    ["Typeset", MathJax.Hub, _this.$buffer[0]],
+                    ['Typeset', MathJax.Hub, _this.$buffer[0]],
                     function(){
+                        //store mathjax "tex", for tex for later mathML conversion
+                        var texJax = _getJaxByElement(_this.$buffer);
 
                         //replace the target element
                         args.target.html(_this.$buffer.html());
 
-                        //store mathjax "tex", for tex for later mathML conversion
-                        var texJax = _getJaxByElement(_this.$buffer);
 
                         //empty buffer;
                         _this.$buffer.empty();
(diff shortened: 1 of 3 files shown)


views/js/qtiCreator/editor/MathEditor.js
```diff
@@ -96,36 +101,42 @@ define(['lodash', 'jquery', 'mathJax'], function(_, $, MathJax){
     MathEditor.prototype.renderFromTex = function(){
 
         var args = _processArguments(this, arguments);
+        var _this = this;
```
krampstudio: Just a notice for later, we agree on a convention for getting the parent scope into a closure. We use `var self  = this;`, except if you're in a worker...

views/js/qtiCreator/test/MathEditor/test.html
```diff
@@ -0,0 +1,29 @@
...
+                require(['taoQtiItem/qtiCreator/test/MathEditor/test'], function(){
+                    QUnit.start();
+                });
+            });
+        </script>
+        <!-- http://xxx/taoQtiItem/views/js/qtiCreator/test/MathEditor/test.html -->
+    </head>
+    <body>
+        <div id="qunit"></div>
+        <div id="qunit-fixture"></div>
```
krampstudio: The fixture element is reset at the end of each test. `QUnit.start` should be called then be called at the end (after all asserts).

views/js/qtiCreator/test/MathEditor/test.js
```diff
@@ -0,0 +1,67 @@
+define([
```
krampstudio: FYI, all tests are supposed to be located under `views/js/test` to be executed by the CI engine. This folder is an old one and will be migrated at some point.
Could you please make the test fails if MathJax is not present, so when we migrate this folder we now why the test do not run.

views/js/qtiCreator/test/MathEditor/test.js
```diff
@@ -0,0 +1,67 @@
...
+    'jquery',
+    'taoQtiItem/qtiCreator/editor/MathEditor'
+],
+function($, MathEditor) {
+    'use strict';
+
+    QUnit.asyncTest('Latex rendering', function test(assert) {
+        var texExprs = [
```
krampstudio: We have a plugin to achieve parameterized testing, you should use it. Take the following test as an example : https://github.com/oat-sa/tao-core/blob/develop/views/js/test/core/encoder/str2array/test.js

views/js/qtiCreator/test/MathEditor/test.js
```diff
@@ -0,0 +1,67 @@
+define([
+    'jquery',
+    'taoQtiItem/qtiCreator/editor/MathEditor'
+],
+function($, MathEditor) {
+    'use strict';
+
+    QUnit.asyncTest('Latex rendering', function test(assert) {
```
krampstudio: We need all tests cases to be wrapped under a module (for reporting), something like : 

`QUnit.module('MathEditor rendering');`

views/js/qtiCreator/editor/MathEditor.js
no-chris: Shall I keep it like this here for consistency with the rest of the file (which is not unit-test covered) ?

views/js/qtiCreator/test/MathEditor/test.js
no-chris: So for now I'm not moving my test as it will be moved with the rest of the folder or shall I move it already ?

krampstudio: no don't move it, it will be migrated. It was just for your information.

views/js/qtiCreator/editor/MathEditor.js
krampstudio: You can let it. The rule regarding code style and old files, is the best effort

PR #587, opened 2016-03-09
Fix/tao 2056 missing tooltip styles dev

krampstudio:
Cherry pick of https://github.com/oat-sa/extension-tao-itemqti/pull/586 to the fix working in develop.
Version upgrade seemed not necessary to reapply.


krampstudio:
wrong branch

PR #600, opened 2016-03-14
added missing minor version

ssipasseuth:
This minor version has been removed during latest merge with develop.

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -211,9 +211,9 @@ class Updater extends \common_ext_ExtensionUpdater
             $this->setVersion('2.14.0');
         }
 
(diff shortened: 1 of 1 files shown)


scripts/update/Updater.php
```diff
@@ -211,7 +211,7 @@ public function update($initialVersion){
             $this->setVersion('2.14.0');
         }
 
-		$this->skip('2.14.0','2.15.0');
+		$this->skip('2.14.0','2.15.0', '2.15.1');
```
krampstudio: Then you should also change the numbers in the block below.

ssipasseuth:
Right, I missused the skip function. Could you check this again please.

PR #601, opened 2016-03-14
fix default ns

ssipasseuth:
Cherry picked https://github.com/oat-sa/extension-tao-itemqti/pull/599
Then we should consider creating a new tag : 2.17.3.


ssipasseuth:
bad PR

PR #612, opened 2016-03-22
Fix/tao 2410 feedback style setting

ssipasseuth:


ssipasseuth:
sorry, mistake in tthe version number in the last commit, it is actually version 2.18.0 to 2.18.1 !

PR #613, opened 2016-03-22
Fix rendering to xml for inline choice interactions

Alroniks:
https://oat-sa.atlassian.net/browse/TAO-2212

### DIFF
--- a/views/js/qtiItem/core/interactions/Interaction.js
+++ b/views/js/qtiItem/core/interactions/Interaction.js
@@ -98,6 +98,11 @@ define([
             _.each(choices, function(choice){
                 if(Element.isA(choice, 'choice')){
                     try{
+                        if (renderer.name == 'xmlRenderer') {
+                            if ((new RegExp('<|>', 'g')).test(choice.text)) {
+                                choice.text = _.escape(choice.text);
+                            }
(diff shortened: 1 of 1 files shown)


views/js/qtiItem/core/interactions/Interaction.js
```diff
@@ -98,6 +98,11 @@ define([
             _.each(choices, function(choice){
                 if(Element.isA(choice, 'choice')){
                     try{
+                        if (renderer.name == 'xmlRenderer') {
+                            if ((new RegExp('<|>', 'g')).test(choice.text)) {
+                                choice.text = _.escape(choice.text);
+                            }
+                        }
```
ssipasseuth: This is not good... 
views/js/qtiItem/core/ is the common qtiItem library and xmlRender is a specific implementation of its renderer.
If that fix need to be done, it should be on the views/js/qtiXmlRenderer

ssipasseuth:
This issue is related to https://oat-sa.atlassian.net/browse/TAO-2284 ...
I will provide the fix for develop as per TAO-2284. I will let you know as soon as I am done with it.
Just prepare to backport it for the client.
Thanks !!

PR #615, opened 2016-03-23
use new itemApi.setVariables() method to prevent race condition

ssipasseuth:
requires https://github.com/oat-sa/extension-tao-item/pull/189


manifest.php
```diff
@@ -29,7 +29,7 @@
     'version' => '2.12.10',
 	'author' => 'Open Assessment Technologies',
 	'requires' => array(
-	    'taoItems' => '>=2.6'
+	    'taoItems' => '>=2.6.7'
```
jsconan: Are you sure it is the right version ? In taoItems I see 2.9.7, not 2.6.7

PR #616, opened 2016-03-23
use new itemApi.setVariables() method to prevent race condition

ssipasseuth:


ssipasseuth:
replaced by https://github.com/oat-sa/extension-tao-itemqti/pull/633

PR #626, opened 2016-03-31
Fix/tao 2460 mathjax error 500

no-chris:
how to test:
- create a new MathML (and not latex) expression with a < or a > operator, for instance

`<mstyle><mrow><mi>x</mi><mo><</mo><mi>y</mi></mrow></mstyle>`

- save the item
- edit the MathML by adding something to the operands (555y for instance)
- save the item
- no error 500 should appear


krampstudio:
Do not forget to backport the fix on develop if the develop branch is also impacted by the bug.

PR #633, opened 2016-04-08
Backport/sync set variables

ssipasseuth:
depends on https://github.com/oat-sa/extension-tao-item/pull/200


ssipasseuth:
replaced by https://github.com/oat-sa/extension-tao-itemqti/pull/633

PR #636, opened 2016-04-11
Allow  'multiple' cardinality for scoring CustomOperator

zagovorichev:


zagovorichev:
@krampstudio  maybe version increase was too much?

scripts/update/Updater.php
```diff
@@ -318,6 +318,7 @@ public function update($initialVersion){
         }
 
         $this->skip('2.20.0', '2.20.1');
+        $this->skip('2.20.1', '2.21.0');
```
krampstudio: only one skip is sufficient : 
```php
$this->skip('2.20.0', '2.20.2');
```

manifest.php
```diff
@@ -26,7 +26,7 @@
     'label' => 'QTI item model',
 	'description' => 'TAO QTI item model',
     'license' => 'GPL-2.0',
-    'version' => '2.20.1',
+    'version' => '2.21.0',
```
krampstudio: you've not added a new feature but fixed one, so the version should be `2.20.2`

zagovorichev:
@krampstudio update

PR #638, opened 2016-04-12
Backport/tao 2323 and tao 2460 mathjax bugs

no-chris:
backport of 
https://github.com/oat-sa/extension-tao-itemqti/pull/606
https://github.com/oat-sa/extension-tao-itemqti/pull/626


krampstudio:
That's good. 

The test is very useful, we should consider adding it to the automated tests at some point (views/js/test) but this would require to automated the install of MathJax. It's another story...

PR #654, opened 2016-04-20
changed color picker position to be close to the color control

no-chris:
this is used by parcc pci like graphFunction or PointAndLine, the color picker should appear close to the mouse cursor and not on top of the item window


ssipasseuth:
oops release-2.12.13 is already tagged https://github.com/oat-sa/extension-tao-itemqti/releases/tag/v2.12.13, I am affraid you need to add this fix to release-2.12.14 (please do not forget the backport to master too)

PR #663, opened 2016-04-27
Fix/color picker position 2.12.14

no-chris:
changed target branch of https://github.com/oat-sa/extension-tao-itemqti/pull/654 to release 2.12.14

backport to develop is here : https://github.com/oat-sa/extension-tao-itemqti/pull/655


ssipasseuth:
I still got the issue ... strangely, if I activate the debugger, the issue seems fixed.
Could it be an issue with race condition ?

PR #664, opened 2016-04-27
Rest Controller to Create,Import and Retrieve Qti Item

absar854:


siwane:
Modifications integrated in TAO.
See PR https://github.com/oat-sa/extension-tao-itemqti/pull/687

PR #671, opened 2016-05-04
use an exclude key to remove a registerd plugin

krampstudio:
This is because we need to perform an exclude instead (see https://github.com/oat-sa/extension-tao-itemqti/pull/658/commits/31dd0cdc8bd913ac7a0dac5e483308ec26333f0e and https://github.com/oat-sa/tao-core/pull/820/files)


krampstudio:
sorry this is wrong!!!

PR #675, opened 2016-05-09
Fix the container finder in the inline feedback autoscroll feature

jsconan:
Take care of the pre-filled `__knownParent__ ` property to secure access to parent window.


krampstudio:
looks ok, @ssipasseuth will take over the review

ssipasseuth:
ok, works for me

PR #677, opened 2016-05-10
Feature/tao 2572 gap match drag and drop

no-chris:
Run the extension update script to create the relevant client lib config registry entries that activates drag & drop

In both gapMatch & graphicGapMatch interactions:
- check that previous behaviour (point & click) is still valid
- check that corresponding actions works now with drag & drop
- in authoring mode, drag and drog handler are not attached as it seems the renderer cannot access the module options (?)


krampstudio:
I've not made the review in the details, so please @ssipasseuth do a more in depth review. The  implementation looks good. 

I have only the following remarks : 

 - the `interact` library should be included (you require it as a Portable Library, which is problematic). This library could even be integrated as a core library (into `tao/views/js/lib`).
 - @ssipasseuth is it the best way for the configuration ?

ssipasseuth:
Hi, can I assume this PR has already been black-box tested and pass all regression test by Artem ?

ssipasseuth:
@krampstudio you are talking about the module configuration for https://github.com/oat-sa/extension-tao-itemqti/pull/677/files#diff-aca42ddeb2f051d43ad25dca11327e2bR114 ?
I think that this is the most straight-forward way to ensure that the behaviour is set on the plateform level, otherwise, we need to add another server side config then ensure that all controller initiating the commonRenderer use that config file... but maybe you have a better suggestion ?
With this method, the options defined in the config becomes the default value but these options may still be overwritten upon instantiation.

views/js/qtiCommonRenderer/renderers/interactions/GapMatchInteraction.js
```diff
@@ -28,8 +28,9 @@ define([
     'tpl!taoQtiItem/qtiCommonRenderer/tpl/interactions/gapMatchInteraction',
     'taoQtiItem/qtiCommonRenderer/helpers/container',
     'taoQtiItem/qtiCommonRenderer/helpers/instructions/instructionManager',
-    'taoQtiItem/qtiCommonRenderer/helpers/PciResponse'
-], function(_, __, $, tpl, containerHelper, instructionMgr, pciResponse){
+    'taoQtiItem/qtiCommonRenderer/helpers/PciResponse',
+    'OAT/interact'
```
ssipasseuth: yes, please call interact from main tao aliase. OAT/interact refers to protable shared library, only to be used within PCIs

views/js/qtiCommonRenderer/renderers/interactions/GapMatchInteraction.js
```diff
@@ -136,35 +148,143 @@ define([
...
+                autoScroll: true,
+                restrict: {
+                    restriction: ".qti-interaction",
+                    endOnly: false,
+                    elementRect: {top: 0, left: 0, bottom: 1, right: 1}
+                }
+            };
+
+            // makes choices draggables
+            interact(choiceSelector).draggable(_.assign({}, dragOptions, {
```
ssipasseuth: maybe you want to use _.default() instead ?

views/js/test/qtiCommonRenderer/interactions/gapMatch/test.html
```diff
@@ -19,6 +19,17 @@
...
             require(['/tao/ClientConfig/config'], function(){
 
+                require.config({
+                    config : {
+                        context : {
+                            'root_url' : '/'
+                        }
+                    },
+                    paths : {
+                        'OAT/interact': '../../../taoQtiItem/views/js/portableSharedLibraries/OAT/interact'
```
ssipasseuth: This may go with https://github.com/oat-sa/tao-core/pull/829 merged

views/js/test/qtiCommonRenderer/interactions/graphicGapMatch/test.html
```diff
@@ -0,0 +1,52 @@
...
+            require(['/tao/ClientConfig/config'], function(){
+
+                require.config({
+                    config : {
+                        context : {
+                            'root_url' : '/'
+                        }
+                    },
+                    paths : {
+                        'OAT/interact': '../../../taoQtiItem/views/js/portableSharedLibraries/OAT/interact'
```
ssipasseuth: This too..

ssipasseuth:
hey, please met me know when it is ready for another round of review :)

PR #691, opened 2016-05-25
Feature/tao 2612 order and associate dragdrop

no-chris:
This adds the drag & drop for the order and associates interaction
And also includes a bit of refactoring and bug fixes for other D&D interactions

Please wait for Artem to review associate interaction before reviewing this


views/js/qtiCommonRenderer/helpers/interactUtils.js
```diff
@@ -0,0 +1,91 @@
...
+ * @author Christope Noël <christophe@taotesting.com>
+ */
+define([
+    'jquery',
+    'lodash',
+    'core/mouseEvent'
+], function($, _, triggerMouseEvent) {
+    'use strict';
+
+    var interactHelper = {
```
ssipasseuth: good idea to create a helper for this, but didn't you initially planned to put this code in tao-core instead ?
it makes sense because the core/dynamicComponent then could benefit from it...

views/js/qtiCommonRenderer/renderers/interactions/AssociateInteraction.js
```diff
@@ -355,38 +351,189 @@ define([
...
+
+            // Drag & drop handlers
+
+            if (self.getOption && self.getOption("enableDragAndDrop") && self.getOption("enableDragAndDrop").associate) {
+                isDragAndDropEnabled = self.getOption("enableDragAndDrop").associate;
+            }
+
+            // Chrome/Safari ugly fix: manually drop element when the mouse leaves the item runner iframe
+            // should be removed when the old test runner is discarded
+            function _iFrameDragFix(draggableSelector, target) {
```
ssipasseuth: Would be a good idea to move all the _iFrameDragFix() and _iFrameDragFixOff() functions into the interact helper.
If variations are needed, could add an optional callback or fire event

ssipasseuth:
orderInteraction : like the drop area hightlight :)
If possible, would be nice to highlight of original container if we drop an element from the right list to the left (original one) to provide an appropirate feedback to user.
Also, when we drag and element from the right list, the box dimensions seem to change a little which gives a little trembling effect. If you have time for it, it would nice to leave the rest of the elements in place when dragging a given element.

ssipasseuth:
associate interaction works great

no-chris:
this PR now depends on https://github.com/oat-sa/tao-core/pull/844

PR #693, opened 2016-05-26
allows any element in infocontrol (even portableInfoControl) in all a…

antoinerobin:
…pip xsd's


bugalot:
Waiting for IMS response about this issue.

PR #695, opened 2016-05-30
move the item title to a dedicated bar

krampstudio:
The new title bar...


ssipasseuth:
humm, I actually find the color of the poll less disturbing.. because now the alternate color looks like a checkerboard which makes the block disaligned and more importantly it is now clearer than the item background which looks as the title is carved into the background.
So what about using the same color as the other header ? And maybe even try a very light shadow at the bottom of the test box to reflect the fact that the title box is above the item background...

PR #708, opened 2016-06-09
Backport 2.31/tao 1910 fix checkbox match interaction

no-chris:
related to https://oat-sa.atlassian.net/browse/TAO-1910

backport of https://github.com/oat-sa/extension-tao-itemqti/pull/707

depends on https://github.com/oat-sa/extension-tao-act/pull/248


ssipasseuth:
This is no longer needed right because it has been released with last sprint final merge, so I guess we close this right ?

PR #710, opened 2016-06-09
Fix/inline feedback autoresize

ssipasseuth:
related to the original bug https://oat-sa.atlassian.net/browse/TAO-2568


no-chris:
shouldn't this fix be on develop also ?

ssipasseuth:
yes, there is another PR for develop https://github.com/oat-sa/extension-tao-itemqti/pull/711

PR #714, opened 2016-06-15
Fix/get item types

antoinerobin:


bugalot:
Tested by importing a QTI package with items (TAO Sample package)

PR #719, opened 2016-06-22
Fix/ghost response creation

ssipasseuth:
This fixes creation of ghost response when the user delete an inline interaction by pressing backspace or by mouse-selection+delete.
This removes a problematic old hack introduced by the fix from http://forge.taotesting.com/issues/3041


krampstudio:
This PR looks good, could you just please add some comments in the source to explain what you're doing (and why), ie. "no crtl-z to prevent interaction deletion", etc.

PR #726, opened 2016-06-30
Support flysystem for item import

jbout:
requires https://github.com/oat-sa/extension-tao-item/pull/216

### DIFF
--- a/model/qti/ImportService.php
+++ b/model/qti/ImportService.php
@@ -113,7 +113,7 @@ class ImportService extends tao_models_classes_GenerisService
      * @throws \common_Exception
      * @return core_kernel_classes_Resource
      */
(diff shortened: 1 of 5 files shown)


model/qti/ImportService.php
```diff
@@ -427,7 +428,7 @@ public function importQtiItem(
                 $rdfItem = $this->createRdfItem((($targetClass !== false) ? $targetClass : $itemClass), $qtiModel);
 
                 $itemAssetManager = new AssetManager();
-                $itemAssetManager->setItemContent($itemService->getItemContent($rdfItem));
+                $itemAssetManager->setItemContent($qtiModel->toXML());
```
bugalot: @jbout, This seems to make the item to be validated two times...

siwane:
Missing apip test

PR #732, opened 2016-07-06
remove $path variable

hutnikau:
After this change:
https://github.com/oat-sa/extension-tao-itemqti/commit/3bffee60f2ac360c259badc559d7f4abc17a23a6#diff-f75b6dac2741b1eca5d33eef5a20e9b7L69
`$path` variable no longer exists.

related PR:
https://github.com/oat-sa/extension-tao-item/pull/220


jbout:
encoders need to be refactored to work with this

PR #733, opened 2016-07-11
Feature/tao 2735 add user modules to new item runner

no-chris:
@krampstudio can you do the code review for the implementation ?

@dietertaotesting can you do the functionnal testing along with https://github.com/oat-sa/extension-tao-act/pull/289 ?

I also created some documentation here : http://forge.taotesting.com/projects/tao/wiki/UserScripts


views/js/runner/provider/manager/userModules.js
```diff
@@ -0,0 +1,56 @@
...
+ * @author Christophe Noël <christophe@taotesting.com>
+ */
+define([
+    'lodash',
+    'module'
+], function (_, module) {
+    'use strict';
+
+    var userModules = [],
+        config = module.config();
```
krampstudio: You should move the call to `module.config()` because the configuration could change. No ?

views/js/runner/provider/manager/userModules.js
```diff
@@ -0,0 +1,56 @@
...
+        load: function load() {
+            return new Promise(function(resolve) {
+                require(userModules, function () {
+                    _.forEach(arguments, function (dependency) {
+                        if (dependency && _.isFunction(dependency.exec)) {
+                            dependency.exec();
+                        }
+                    });
+                    resolve();
+                });
```
krampstudio: add a rejection (or even a resolve if we'd like to ignore loading errors) as the 2nd argument of  `require`.

views/js/runner/provider/manager/userModules.js
```diff
@@ -0,0 +1,56 @@
...
+
+    var userModules = [],
+        config = module.config();
+
+    if (config && config.userModules && _.isArray(config.userModules)) {
+        userModules = config.userModules;
+    }
+
+    return {
+        load: function load() {
```
krampstudio: Please add a jsdoc comment

views/js/runner/provider/qti.js
```diff
@@ -158,7 +159,14 @@ define([
                          */
                         self.trigger('listpic', picManager.collection(self._item));
 
-                        done();
+
+                        userModules.load()
```
krampstudio: In that context, you can chain the promises, so you don't need to repeat the `catch` clause.

```js
return userModules.load().then(done);
```

views/js/runtime/qtiBootstrap.js
```diff
@@ -60,35 +61,38 @@ define([
...
+
+                                iframeNotifier.parent('itemloaded');
+
+                                //IE9/10 loose the iframe focus, so we force getting it back.
+                                _.defer(function(){
+                                    window.focus();
                                 });
+                            })
+                            .catch(function(err) {
+                                throw new Error('error', 'Error in user modules : ' + err.message);
```
krampstudio: the 1st arg of `Error` is the message

krampstudio:
This is great but you should restructure the module. Instead of having the list of modules in the module context, it would be better to load them on demand, so you won't need the `setUserModule` anymore.


Something like : 
```js

define([
    'lodash',
    'module'
], function (_, module) {
    'use strict';

    return {
        load: function load(userModules ) {
          var config = module.config();
          userModules = userModules || [];
          if(config && _.isArray(config.userModules) && config.userModules.length){
              userModules = userModules.concat(config.userModules);
          }

            return new Promise(function(resolve) {
                require(userModules, function () {
                    _.forEach(arguments, function (dependency) {
                        if (dependency && _.isFunction(dependency.exec)) {
                            dependency.exec();
                        }
                    });
                    resolve();
                });
            });
        }
    };
});
```

no-chris:
Removing setUserModule makes it difficult to test the integration of user modules in the item runner.

krampstudio:
If the purpose is the testing, you can change the config during the tests, using: 

```js
requirejs.config({
  config : {
      'taoQtiItemrunner/provider/manager/userModules' : []
  }
});
```

no-chris:
@krampstudio you can review again

PR #736, opened 2016-07-13
just prevent unwanted typeerror

krampstudio:
I think I've got the error when being timed out on an item that have a modal feedback. But this patch will prevent a nasty `TypeError` at least.


ssipasseuth:
Yes, this additional verification is a good addition .
However I find strange is that the feedback attributes is always empty even when its attributes are supposed to have been loaded, i.e: feedback display condition fullfillled so this patch is potentially hiding potential development bug in some cases.
After a debug, I found out that feedbackData is a string, so you certainly need a JSON.parse(feedbackData) before passing the returned object to loadElements right ?

ssipasseuth:
@jsconan please take over this small PR while I am away, thanks !

PR #749, opened 2016-08-08
Added extended description on UnsupportedQtiType exception

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-448


model/qti/ParserFactory.php
```diff
@@ -1455,29 +1455,29 @@ private function buildFeedback(DOMElement $data, $interactive = false){
...
+            /*if($interactive){
                 $container = new ContainerFeedbackInteractive();
                 $myFeedback = new $feedbackClass($container, $attributes, $this->item);
                 $this->parseContainerInteractive($data, $myFeedback->getBody());
             }else{
                 $container = new ContainerStatic();
                 $myFeedback = new $feedbackClass($container, $attributes, $this->item);
                 $this->parseContainerStatic($data, $myFeedback->getBody());
-            }
+            }*/
```
ssipasseuth: I think we might as well as remove this comment codes. It is not pretty to release codes with this.
We want to recover this code by searching the commit history. If you want to make this simply, you may use clear key words such as "removed feedbackInline parsing" etc. what do you think of that?

zagovorichev: I'm totally agree with you

zagovorichev:
@ssipasseuth Update

PR #764, opened 2016-08-23
fixed the regex for mime type parsing

ssipasseuth:


jsconan:
Still have issue with the MP4 Video MIME type. Maybe could you also add support for digits in your regex ?

PR #766, opened 2016-08-25
Merge develop/portable migration

ssipasseuth:
**This new PR replaced the old one https://github.com/oat-sa/extension-tao-itemqti/pull/765**

This PR reimplements the whole life cycle for portable elements (PCI+PIC) according to https://oat-sa.atlassian.net/browse/TAO-2524

It involves the import/export of a portable element package (for authoring and runtime), the loading of portable element during authoring, preview and delivery.

Portable elements now indeed need to be registered on a platform level before being able to be rendered or authored.
The PCI sources are now loaded from a global central location (instead of from a local item director).

From this main PR depends the other extensions that implement portable elements 
oat-sa/extension-tao-itemqti-pci#38
oat-sa/extension-tao-itemqti-pic#25
oat-sa/extension-pcisample#10
oat-sa/extension-tao-mp#241

This is mostly a code review, the functional testing is being taking care of by the QA team, which bases their tests on the AC and the doc https://docs.google.com/document/d/1rtwPSUkSaiCHhaIbpc4CXEQhca8D6Z1pH9ciKvO_gsM/edit?usp=sharing

For server-side related questions, please address them to @siwane.
For client-side codes, please ask @ssipasseuth.

The main reviewer is @jbout who has the final word on this PR.
@jsconan please review the client side (registry plus loading mechanism)

PR #773, opened 2016-09-05
Fix/portable migration

chrisphpgarcia:
fix and reaplce https://github.com/oat-sa/extension-tao-itemqti/pull/766


siwane:
Will be replace by branch https://github.com/oat-sa/extension-tao-itemqti/tree/feature/portableElement

PR #779, opened 2016-09-07
Fix/better logging for qti validation

boajer:
Validation errors gets only in logs with notice level, that makes diagnostic by logs more complicated.


model/qti/parser/ValidationException.php
```diff
@@ -43,4 +43,10 @@ public function getReport()
     {
         return common_report_Report::createFailure(__("Malformed XML:\n%s", implode("\n", $this->errors)));
     }
+
+    public function getSeverity()
+    {
+        return \common_Logger::ERROR_LEVEL;
```
hutnikau: In my opinion `WARNING_LEVEL` will be more appropriate here because errors in QTI XML file found by validator are in fact not the system level errors.

PR #780, opened 2016-09-08
Feature/tao 2169 qti object authoring

ssipasseuth:
requires https://github.com/oat-sa/tao-core/pull/983


views/js/qtiCreator/widgets/static/object/states/Active.js
```diff
@@ -13,188 +13,120 @@
...
-                _initAdvanced(_widget);
+                refreshRendering(_widget);
             },
-            alt : formElement.getAttributeChangeCallback(),
-            longdesc : formElement.getAttributeChangeCallback(),
-            align : function(media, value){
-                inlineHelper.positionFloat(_widget, value);
+            width : function(object, value){
+                var val = parseInt(value, 10);
+                if(_.isNaN(value)){
```
no-chris: caution with _.isNaN:
_.isNaN('text') = false
you may want to use native isNaN with caution also, but _.isFinite might be what you are looking for (to double checked)

ssipasseuth: well the values comes from parseInt() which returns either a valid integer of NaN anyway so such a situation should not happen right ?
https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Objets_globaux/parseInt

no-chris: then you need 
var val = parseInt(value, 10);
+                if(_.isNaN(**val**)){

and not 
var val = parseInt(value, 10);
+                if(_.isNaN(**value**)){

PR #784, opened 2016-09-12
Fix/packer refactoring

boajer:


model/Export/AbstractQTIItemExporter.php
```diff
@@ -173,7 +172,7 @@ protected function getStorageDirectory(\core_kernel_classes_Resource $item, $lan
 
         //we should use be language unaware for storage manipulation
         $path = str_replace($lang, '', $directory->getPrefix());
-        $storageDirectory = new \tao_models_classes_service_StorageDirectory($item->getUri(), $directory->getFilesystem()->getId(), $path);
+        $storageDirectory = new Directory($directory->getFilesystem()->getId(), $path);
         $storageDirectory->setServiceLocator($this->getServiceManager());
 
         return $storageDirectory;
```
hutnikau: I would recommend use this approach:
```php
protected function getStorageDirectory(\core_kernel_classes_Resource $item, $lang)
    {
        $itemService = \taoItems_models_classes_ItemsService::singleton();
        $directory = $itemService->getItemDirectory($item, $lang);
        //we should use be language unaware for storage manipulation
        $path = str_replace($lang, '', $directory->getPrefix());
        return $itemService->getDefaultItemDirectory()->getDirectory($path);
    }
```

PR #785, opened 2016-09-14
fixed rubrick block rendering when it should not displayed

ssipasseuth:


krampstudio:
Could you please describe the original issue. Is it when an item did contain a `<rubricBlock>` element but without a body ?

ssipasseuth:
Yes, sorry. That is right. To test it you can import and item containing an empty rubricBlock or a rubrickBlock that has some content but not the right role for the candidate, e.g. author.

krampstudio:
I've tried with the following [item](https://github.com/oat-sa/extension-tao-itemqti/blob/master/test/samples/xml/qtiv2p1/rubricBlock/extended_text_rubric.xml)
it was rendered correctlly without your PR in the preview (and all the item is blank in the authoring). With your PR, I've a blank screen (no error in the web console) for both preview and authoring :s

ssipasseuth:
strange ... you should have no difference in authoring (since the authoring displays rubrick block no matter the role given to it) but there is a visible empty div in the preview without the fix. Please check the screenshots below.

Authoring in both branches (with develop merged in):
authoring_both

Preview on develop
preview_before_fix

Preview with the fix (develop merged in)
preview_with_fix

Could you please check again and let me know if you still have the issue you mentioned... thanks !

krampstudio:
I have imported again the item and it seems to work now...

PR #794, opened 2016-09-20
made media insertion optional in container editor

ssipasseuth:
taoQtiItem/qtiCreator/editor/containerEditor is used by PCI creator widgets to edit their html content. However PCI does not handle object element natively so it is wise to made this optional and deactivate it by default. This PR adds this option.


krampstudio:
I've tested and I don't have the "inset media" option inside a PCI content.

PR #796, opened 2016-09-20
Always Iterating array

boajer:


siwane:
Hello,

I already did a fix assigned to Joel.

PR #797, opened 2016-09-21
fix empty nodes

siwane:


antoinerobin:
duplicate : https://github.com/oat-sa/extension-tao-itemqti/pull/796

PR #802, opened 2016-09-23
Tao 3126 fix graphic gap match interaction correct response

hutnikau:
I updated `oat\taoQtiItem\model\update\ItemUpdater` class to enable it work with flysystem. Now string or `oat\oatbox\filesystem\Directory` can be passed to `__construct` method. @ssipasseuth, please review it.

@krampstudio, please check `oat\taoQtiItem\model\update\FixHotspotCorrectResponse` class which is fix of TAO-3126 issue.


krampstudio:
we need to wait my PR to merge this one to prevent creating a possible system inconsistency.

ssipasseuth (approved):
ok works well. tested this way: 

# item data with one broken item (develop) and one fixed item by https://github.com/oat-sa/extension-tao-itemqti/pull/809.

# the script correct fixes the broken one as expected and the correct ones is unchanged.

ssipasseuth:
(oops sorry for the huge font, I am confusing the ordered list syntax of md with jira ...)

ssipasseuth:
So it appears this PR along with the fix #809 are good. 
Since it is our rule not include item modification in updater, the question now is how we are going to proceed with the fix of existing items in existing instance. 
@llecaque any suggestion ?

krampstudio:
Closing it for now. We will rely on a configuration option instead for the backward compat.

PR #816, opened 2016-10-03
Feature/tao 3140 get item categories

antoinerobin:


test/ItemCategoriesServiceTest.php
```diff
@@ -0,0 +1,137 @@
...
+            'itemUri1' => array('SCIENCE', 'MATH', 'MEDIUM'),
+            'itemUri5' => array('ENGLISH', 'HIGH'),
+            'itemUri3' => array('MATH', 'LOW'),
+            'itemUri4' => array(),
+            'itemUri2' => array('ENGLISH'),
+        );
+        /** @var ItemCategoriesService $itemCategoriesService */
+        $itemCategoriesService = new ItemCategoriesService(
+            array(
+                'properties'    => array('subject', 'difficulty'),
```
bugalot: I would rather suggest to bind a mapping to a single property rather than multiple, to improve isolation and possible side effects on large mappings.

e.g. 

```php
array(
    'http://mysite.com#prop1' =>
    array(
        'mysite.com#instance1' => 'MATH',
        'français' => 'FRENCH'
    ),
    'http://mysite.com#prop2' => 
    array(
        'doutchlund' => 'GERMAN'
    )
)
```

test/ItemCategoriesServiceTest.php
```diff
@@ -0,0 +1,137 @@
...
+                    'hard' => 'HIGH'
+                    )
+            )
+        );
+        $categories = $itemCategoriesService->getCategories($items);
+
+        $this->assertEquals($expected, $categories);
+    }
+
+}
```
bugalot: Could it be possible to also have unit tests object resources to categories e.g. 'http://mysite.com#prop1' => 'GLOUGLOU' ?

PR #818, opened 2016-10-03
fix QtiOutput unit test

hutnikau:
After this commit:
https://github.com/oat-sa/extension-tao-itemqti/commit/2ac0b0bd63e64d1db4ceb60a4b405e29032fd606
`QtiOutputTest` unit test was broken.


ssipasseuth:
This indeed fixes QtiOutputTest, thanks !

PR #830, opened 2016-10-14
Feature/register individual portable element

ssipasseuth:
I have added in this PR an abstract invokable to allow easy registration from individual portable element directory source (PCI or PIC).


model/portableElement/action/RegisterPortableElement.php
```diff
@@ -0,0 +1,90 @@
...
+        if (!is_readable($sourceDirectory)) {
+            return $this->createFailure('the source directory does not exists or is not readable ' .$sourceDirectory);
+        }
+
+        try {
+            $model = $service->getValidPortableElementFromDirectorySource($sourceDirectory);
+            if(!empty($params)){
+                $minRequiredVersion = $params[0];
+                // if the minimal required version number string "x.y.z" is given in the parameter, the new target version should be equal or higher than it
+                if(version_compare($model->getVersion(), $minRequiredVersion) < 0){
```
ssipasseuth: It is slightly different here as it will compare the version in the source directory with the minimal required defined in the parameter. This (optional) param allows setting a required minimal version as in this example : https://github.com/oat-sa/extension-pcisample/pull/20/files#diff-53f22eb141a5b2bf0797704363c4aa12R52

PR #839, opened 2016-11-03
Fix/tao 3251 item responses

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3251

Fix the behaviour of the shortcut that allows to check a choice (SPACE key).

Without this fix, when you press the space bar to check a response, the radio button is checked, then immediately unchecked

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
@@ -62,6 +62,7 @@ define([
             }
 
             if( keyCode === KEY_CODE_SPACE || keyCode === KEY_CODE_ENTER){
+                // delay the trigger to be sure the selection will not be invalidated by the browser
                 _.delay(function(){
                     _triggerInput($this.closest('.qti-choice'));
                 }, 100);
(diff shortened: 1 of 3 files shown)


views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
```diff
@@ -55,13 +55,16 @@ define([
...
         $choiceInputs.on('keydown.commonRenderer', function(e){
+            var $this = $(this);
             var keyCode = e.keyCode ? e.keyCode : e.charCode;
             if(keyCode !== KEY_CODE_TAB){
                 e.preventDefault();
             }
 
             if( keyCode === KEY_CODE_SPACE || keyCode === KEY_CODE_ENTER){
-                _triggerInput($(this).closest('.qti-choice'));
+                _.delay(function(){
```
ssipasseuth: might be useful to add a comment for future developer who will see this code

jsconan: This PR  is only intended to fix the `space` key selection. The tab navigation has been made inside another PR, related to the test runner only.

jsconan:
This PR  is only intended to fix the `space` key selection. The tab navigation has been made inside another PR, related to the test runner only.

views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
jsconan: Sure!

ssipasseuth:
This is good enough for now but does not solve the real issue. The real issue is that somehow the click event callback is executed when the focus is set on an input and space key is pressed. We need to investigate this further.

PR #845, opened 2016-11-22
Fix/pci properties entry replacement

ssipasseuth:
Fixed compilation and delivery for pcis which have any property entry htmlentity-encoded .

@no-chris please validate the functionality and check that fix is applicable to the client who reported it
@siwane please review the php code, this solution fixes a the nodeValue replacement, in case it is html-encoded. Somehow such DomNode when given an htmlentity-encoded value is decoded when one tries to access it with $mode->nodeValue.


no-chris:
This fixes the client problem

PR #846, opened 2016-11-24
Feature/delimiter configurable csv

siwane:


model/flyExporter/simpleExporter/ItemExporter.php
```diff
@@ -45,7 +45,7 @@ class ItemExporter extends ConfigurableService implements SimpleExporter
     const CSV_DELIMITER = ',';
 
     /**
-     * Defautl csv enclosure
+     * Defautlt csv enclosure
```
antoinerobin: try again, it's almost that

PR #850, opened 2016-11-28
refer prototype branch.

bugalot:
JIRA: https://oat-sa.atlassian.net/browse/TAO-3410

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
     },
     "minimum-stability" : "dev",
     "require": {
-        "oat-sa/lib-tao-qti" : "~2.0.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -56,7 +56,7 @@
     },
     "minimum-stability" : "dev",
     "require": {
-        "oat-sa/lib-tao-qti" : "~2.0.0",
+        "oat-sa/lib-tao-qti" : "dev-feature/TAO-3410-flysystem-test-state-files",
```
bugalot: After experimentation, will have to refer to "oat-sa/lib-tao-qti":"~2.1.0" before merging.

PR #851, opened 2016-11-30
Feature/tao 3370 limit input length

ssipasseuth:
extract the existing pattern mask management from extended text interaction and apply it to both string interactions.

works better with https://github.com/oat-sa/tao-core/pull/1085 : the tooltip will display better.


views/js/qtiCommonRenderer/tpl/interactions/textEntryInteraction-counter.tpl
```diff
@@ -0,0 +1,6 @@
+<span class="qti-textEntryInteraction-counter">
```
no-chris: is this still used ?

ssipasseuth: no indeed, nice catch !

views/js/test/qtiCommonRenderer/interactions/textEntry/test.js
```diff
@@ -0,0 +1,192 @@
...
+                assert.equal(getTooltipContent($input), '5/5', 'the warning message is correct');
+                assert.ok(getTooltip($input).is(':visible'), 'warning tooltip is visible');
+                assert.ok($input.hasClass('maxed'), 'has state maxed');
+
+                $input.val('1234');
+                $input.keyup();
+                assert.equal(getTooltipContent($input), '4/5', 'the instruction message is correct');
+                assert.ok(getTooltip($input).is(':visible'), 'info tooltip is visible');
+                assert.ok(!$input.hasClass('maxed'), 'has state maxed removed');
+
```
ssipasseuth: Interesting yes, but unfortunately, I didn't find a way to do this. Indeed if we simulate a key input event in an input, the event is fired but the input value is not updated so no way to really simulate a use typing one by one, which is a pity.

ssipasseuth:
Ok, changed applied, please check again.

PR #858, opened 2016-12-09
 Sesstion Stickyness FileUpload support

boajer:
depends on https://github.com/oat-sa/tao-core/pull/1104


model/import/QtiPackageImport.php
```diff
@@ -39,71 +40,71 @@
...
-		} else {
-		   throw new common_exception_Error('No source file for import');
-		}
-		return $report;
+            $uploadedFile = $this->getLocalCopy($fileInfo['uploaded_file']);
+
+            //test versioning
+            $repository = null;
+            if (common_Utils::isUri($form->getValue('repository'))) {
+                $repository = new core_kernel_versioning_Repository($form->getValue('repository'));
```
siwane: Please can u replace core_kernel_versioning_Repository with FileSerializer?

PR #871, opened 2017-01-16
Fix/tao 3626 dont propagate choice navi

krampstudio:
On a choice interaction, with the themeSwitcher plugin and shortcuts activated : 
 - hit `tab`
 - open the theme switcher
 - play with  :arrow_down:, :arrow_up: and  :leftwards_arrow_with_hook:

The main issue was the incompatibility of the native event's `preventDefault` of `util/shortcut/registry` with jQuery's events of the render: you can prevent native events from jQuery handlers but not the other way. So I've prevented the bubbling into the renderer (those events  didn't have to bubble anyway).

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -29,7 +29,7 @@ return array(
     'label' => 'QTI item model',
     'description' => 'TAO QTI item model',
     'license' => 'GPL-2.0',
-    'version' => '6.8.6',
+    'version' => '6.8.7',
     'author' => 'Open Assessment Technologies',
     'requires' => array(
         'taoItems' => '>=2.25.0',
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -446,7 +446,7 @@ class Updater extends \common_ext_ExtensionUpdater
             $this->setVersion('6.8.2');
         }
 
-        $this->skip('6.8.2', '6.8.6');
+        $this->skip('6.8.2', '6.8.7');
     }
 
 }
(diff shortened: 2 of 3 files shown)


krampstudio:
I've reformatted the whole file. This [diff](https://github.com/oat-sa/extension-tao-itemqti/pull/871/commits/495ca398a14fcd83159a29b19a90ba22125ef306) is more readable...

no-chris:
Is this to be backported as well?

no-chris:
Maybe we should prevent all other shortcuts when the theme switcher is opened ? 
One could argue that the following key sequences are still unexpected behavior:
- open the switcher
- hit tab
- I would expect that it would not do anything, and all shortcuts to be scoped on the theme switcher until it is closed, but here it focus on the item behind

I don't know if this is usable for this purpose : https://github.com/oat-sa/tao-core/blob/master/views/js/util/shortcut/registry.js#L782-L784
And by registering up/down/enter on another registry instance ?
I remember JS doing something similar for the dialog box
https://github.com/oat-sa/extension-tao-testqti/pull/673/files

krampstudio:
I can close the TS if it loose the focus. Disabling all shortcuts seems a bit extreme (dialogs are modal so it is expected to only concentrate on them).

> And by registering up/down/enter on another registry instance ?

This will lead to many issues since the registry is global : what about multiple interactions ? conflicts with default behavior (up & down for scrolling) ?  I was more thinking to remove them from the registry and catch them using events bound to the nodes. 

I'll come back with something.

krampstudio:
I'll create backports once done

krampstudio:
I've updated the companion PR https://github.com/oat-sa/extension-tao-testqti/issues/707

krampstudio:
A backport was not necessary but a release for this version : https://github.com/oat-sa/extension-tao-itemqti/releases/tag/v6.8.7

PR #874, opened 2017-01-19
TAO-3670 Added custom interaction key and logic to figure out the typ…

gyszucs:
…e of the custom interaction.


antoinerobin:
could you please add a version bump ?

PR #880, opened 2017-01-27
Scale graphic point according to transform as well

eacraft13:
TAO-3257

### DIFF
--- a/views/js/qtiCommonRenderer/helpers/Graphic.js
+++ b/views/js/qtiCommonRenderer/helpers/Graphic.js
@@ -762,36 +762,24 @@ define([
 
             //get the click coords
             var point = this.clickPoint($container, event);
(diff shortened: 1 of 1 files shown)


views/js/qtiCommonRenderer/helpers/Graphic.js
```diff
@@ -781,6 +781,19 @@ define([
                 }
             }
 
+            // recalculate again in case of scaled image via transform
+            var matrix = $container.closest('.transform-scale').css('transform');
```
krampstudio: Please consider using the module `ui/transformer`to get the matrix values, it manage the vendor prefix also because there still browsers that use `-webkit-transform`

eacraft13:
This was an in progress solution and was not merged for a better solution.

PR #894, opened 2017-02-21
Fix/tao 3898 disbale elements

dietertaotesting:
Depencies between:
https://github.com/oat-sa/extension-tao-testqti/pull/756
https://github.com/oat-sa/extension-tao-itemqti/pull/894
https://github.com/oat-sa/tao-core/pull/1199
https://github.com/oat-sa/extension-tao-act/pull/493, instructions are provided there


views/js/qtiCommonRenderer/renderers/interactions/ChoiceInteraction.js
```diff
@@ -103,6 +103,7 @@ define([
             var $choiceBox = $(this);
             var state;
             var eliminator = e.target.dataset && e.target.dataset.eliminable;
+            var input = this.querySelector('.real-label > input');
```
jsconan: Why don't you use jQuery to access the DOM? Is there a particular reason?
I am asking that because the other code in the same module is using jQuery

dietertaotesting: I try to avoid jQuery when it just bloats the code whitout any added value

PR #895, opened 2017-02-24
Ensure creation of paper and image to prevent TypeError

eacraft13:
TAO-3257

@hutnikau, I believe this should solve the issues you discovered. I would appreciate if you could verify. If not, perhaps you could send me a test item (or whatever) so that I could duplicate the issue.


hutnikau:
@eacraft13, looks good for me. At least I don't get errors anymore. Thanks for fast fix.

krampstudio:
Please increase version number in both `manifest.php` and `scripts/update/Updater.php`

PR #899, opened 2017-03-03
Feature/tao 3294 save before leaving

krampstudio:
In order to achieve the given goal : "don't loose data when leaving the item creator", I had to refactor the mechanism that track item changes.
I've introduced a new plugin `changeTracker` that check if the current item has changed and hook the following actions : 
 - display the preview.
 - leave the authoring in TAO using the back button.
 - leave the authoring by clicking one of the link in the header.
 - leave the authoring by closing the browser (here we have the native popup, since the `beforeunload`event provides only this mechanism for obvious reasons)
 
Each action but the last displays a confirmation message only if there are changes. If you save your changes the confirmation doesn't trigger.

The only mechanism that isn't implemented is using the browser back button.


views/js/qtiCreator/plugins/content/changeTracker.js
```diff
@@ -0,0 +1,205 @@
...
+                        })
+                        .trigger('save', true);
+                });
+            };
+
+            //are we doing the confirm process ?
+            var asking = false;
+
+            /**
+             * Display a confirmation dialog, "ok" means
```
krampstudio: I forgot... :/

PR #901, opened 2017-03-07
Fix/tao 3263 graphic gap match

eacraft13:


eacraft13:
@dietertaotesting  & @jsconan, I created a new branch and pr for tao-3263. I let the old one get a bit dirty and wanted to make it easier to review.

views/js/qtiCreator/widgets/interactions/graphicGapMatchInteraction/states/Question.js
```diff
@@ -47,30 +47,32 @@ define([
...
-        params.$target.css({
-            width: params.width * factor,
-            height: params.height * factor
-        })
-            .attr('width', params.width * factor)
-            .attr('height', params.height * factor);
+        params.$target
+            .css({
+                width: params.width * factor,
+                height: params.height * factor
```
jsconan: I don't understand the purpose here: your removed the jsdoc about `factor`, but left it in the function's params. Then you still use it in the CSS properties (without adjustment if undefined), but not in the attributes... Isn't there an inconsistency?

eacraft13: Yes, this method needed better documentation. Ultimately, maintaining different widths between style and attributes was needed to be able to maintain proper scale on resize events (proper resizing also being addressed in this pull request).

Two more things to mention: 

1) This code was created as a fix, as opposed to a rewrite (I thought a rewrite was out of scope but I do recognize that the original code could be redone more intuitively). 

And 2) the first commit (that removed factor) was to make the UX more intuitive (e.g., adjusting the mediasizer to 100px produced a 100px gapImg, not a 100px * factor gapImg). This was later adjusted (back to 100px * factor gapImg) to be more consistent with current code (this is how the hotspots and correct states were created).

PR #913, opened 2017-04-10
Fix/tao 4121 media print

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-4121

Requires: https://github.com/oat-sa/extension-tao-item/pull/262

Fix the renderers manager.
Make the class `QtiItemPacker` more modular and more overridable.


siwane:
@ssipasseuth, it's for me

PR #923, opened 2017-05-01
Fixed incorrect behavior of the saving confirmation window

zagovorichev:


views/js/qtiCreator/plugins/content/changeTracker.js
```diff
@@ -220,7 +223,11 @@ define([
                     styleChanges = false;
                 })
                 .before('exit', function(){
-                    return confirmBefore(messages.exit);
+                    return confirmBefore(messages.exit)
+                        .then(function(){
+                            removeChangeHandlers();
+                        })
+                        .catch(_.noop); //do nothing but prevent uncatched error
```
krampstudio: The `catch(noop)` at this position isn't necessary since the promise is returned (so the `catch` will be handled at a higher level).

zagovorichev:
@krampstudio Update

PR #926, opened 2017-05-08
switched from portable Token websource to Action websource

ssipasseuth:
After discussing with @jbout, it has been agreed to change the websource of portable elements from Token to Action. So from now on, PCI sources no longer suffer a TTL and they are accessible as longer there is a TAO session opened. PCI sources are definitely considered as source code.


krampstudio:
Not sure if it's related to the new web source but this PCI has something not loaded : 

thumb

I'll investigate.

ssipasseuth:
You are right, it is related to the websource switch. The ActionWebsource use another file mimetype detection and it returns sometimes svg file as text/plain instead of image/svg+xml.
I will look for a solution.

ssipasseuth:
So I have created another PR that should address this issue: https://github.com/oat-sa/tao-core/pull/1312

PR #932, opened 2017-05-18
Fix/tao 3598 mediaplayer can seek

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3598

Requires: https://github.com/oat-sa/tao-core/pull/1333

When the `maxPlays` option is set, the Media Interaction does not allow to seek a position within the media using the duration bar.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -29,11 +29,11 @@ return array(
     'label'       => 'QTI item model',
     'description' => 'TAO QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '8.4.1',
+    'version'     => '8.4.2',
     'author'      => 'Open Assessment Technologies',
     'requires' => array(
         'taoItems' => '>=4.1.0',
-        'tao'      => '>=10.6.0'
+        'tao'      => '>=10.7.0'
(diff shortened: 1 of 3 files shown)


ssipasseuth:
No problem with the code.
However the 2nd AC of the issue TAO-3598 is not fulfilled: "when max plays is set, playing/pausing should be allowed (current behavior)" -> could you please double check this ?

jsconan:
Ok, did not pay attention to that point, as it is mentioned this is the current behavior.
I added the check.

However, I am wondering about this hidden feature: the "pause" option is not set (the checkbox is still unchecked), but the behavior is here. Should not we also disable the option and force it to enabled when max play is set?

ssipasseuth:
Humm I think you are right. The pause function is not standard QTI and has been added to give us an extra configuration option. But indeed since this option is separate from maxChoice maybe that AC simply means that the pause/not pause feature should be compatible with maxPlays. Therefore I think your initial implementation is correct.
Maybe there is an misunderstanding here. So maybe you should clarify this point with the product owner -> Mark.

PR #933, opened 2017-05-23
Fix/tao 3097 map response properties

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-3097

Disable the MapResponse form when no response mapping is set, as the properties won't be saved without mapping.


ssipasseuth:
Hi, no problem from the code point of view.
But I am more concerned about the user experience as the disabled inputs are not very visible and most importantly, an _innocent_ user doesn't know why it is initially disabled. So I would suggest improving the visual feedback to the user. 
Two possibilities I see (there may be more).
1. Try making the disabled section of the form more obvious, not only the inputs but also play maybe with the background color or the text color.
2. Hide the input+label completely (while keeping the header )
In all cases, I believe it would be beneficial to add an info message at the beginning of the form with a short explanation like "mapping options are available when at least one map entry is defined"

PR #946, opened 2017-06-07
Fix/tao 4438 mathjax rendering time

no-chris:
https://oat-sa.atlassian.net/browse/TAO-4438

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/Math.js
+++ b/views/js/qtiCommonRenderer/renderers/Math.js
@@ -33,6 +33,7 @@ define([
     'use strict';
 
     // Do not wait between rendering each individual math element
+    // http://docs.mathjax.org/en/latest/api/hub.html
     MathJax.Hub.processSectionDelay = 0;
 
     return {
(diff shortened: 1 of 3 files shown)


views/js/qtiCommonRenderer/renderers/Math.js
```diff
@@ -32,22 +32,26 @@ define([
 ], function(_, Promise, tpl, containerHelper, MathJax){
     'use strict';
 
+    // Do not wait between rendering each individual math element
+    // http://docs.mathjax.org/en/latest/api/hub.html
+    MathJax.Hub.processSectionDelay = 0;
```
krampstudio: WTF ? Why is there an option to delay between renderings ? Anyway...
Please move the configuration in the render function itself or wrap it around 
```
if(typeof(MathJax) !== 'undefined' && MathJax){
```
because `MathJax` might be the fallback (if it's not installed, we will have a `TypeError call Hub on undefined`)

no-chris:
requested change done!

PR #947, opened 2017-06-08
tao-4441 Answer eliminator updates for MathML and graphic items

eacraft13:
[tao-4441](https://oat-sa.atlassian.net/browse/TAO-4441)

This adds a span tag around item content so that eliminated items can most easily set a background (e.g. a x'ed out pattern) according to the item's content width.


views/js/qtiCommonRenderer/tpl/container.tpl
```diff
@@ -1,5 +1,5 @@
 {{~#equal contentModel "blockStatic"~}}
-<div class="qti-block">{{{body}}}</div>
+<div class="qti-block"><span>{{{body}}}</span></div>
```
ssipasseuth: could we not simply rely on the <div class="qti-block"> to enable the styling ? 
This div has indeed been added to do such a thing.

eacraft13: Yes, a span was a bad idea.

eacraft13:
@ssipasseuth, I've updated to remove the span. However, I still need an extra element to apply the 'x' to. This is because I'm now using an svg to create the 'x'. Ideally, I'd be able to do this with a css background (such as the linear-gradient I had originally tried). The problem I was having is that I can't figure out how to get it to reliably show. Sometimes it showed with width 100%, but most of the time you would need to scale down the width percentage to get it to show (which meant the 'x' would not cover the item properly anyway). Perhaps you would know how to do so? But I'm properly stumped as to why.

views/js/qtiCommonRenderer/tpl/container.tpl
```diff
@@ -1,5 +1,9 @@
 {{~#equal contentModel "blockStatic"~}}
 <div class="qti-block">{{{body}}}</div>
+<svg class="qti-block-overlay">
+    <line x1="0" y1="100%" x2="100%" y2="0" style="stroke:rgb(0,0,0);stroke-width:1"/>
+    <line x1="0" y1="0" x2="100%" y2="100%" style="stroke:rgb(0,0,0);stroke-width:1"/>
+</svg>
```
ssipasseuth: if this is specific to answer eliminating functionality, I would not put it in the common container.tpl otherwise all blockstatic QTI element will have this div as sibling although they are not using it. So I would recommend moving this directly to the answer elimitor functionality in the choice interaction itself. 
In this process, you could probably rename the qti-block-overlay to something more specific to answer eliminating if this overlay is only supposed to be used for answer elimination.

ssipasseuth:
integrated the style from https://github.com/oat-sa/extension-tao-act/pull/574

PR #951, opened 2017-06-12
Fix/tao 4441 apply elimination svg to length of the item

eacraft13:


eacraft13:
@ssipasseuth, the added styles allows for the 'x' to be the length of the item only. As opposed to an 'x' that takes up the whole div's length. The img's **style width** of 100% is so the img's **attribute width** doesn't get applied twice when using display of inline-block.

Perhaps you preferred the 'x' to take up the whole width of the div. If so, then you can close this pr without merging.

views/scss/qti/_choice.scss
```diff
@@ -40,6 +40,14 @@
...
         .choice-area {
             margin-top: 25px;
+            .label-content {
+                display: inline-block;
+                .qti-block {
+                    img {
+                        width: 100%;
+                    }
+                }
+            }
```
ssipasseuth: oops, I thought it is already working this way: so maybe the style I mistakenly integrated on Friday on your behalf should be removed here? https://github.com/oat-sa/extension-tao-itemqti/blob/master/views/scss/qti/_choice.scss
What do you think ?

ssipasseuth: On top of that I found another potential issue: you can normally define the size in percentage of a picture in the QTI creator (in the property form on the right). But now even if you set a percentage, e.g. 25%, it always show 100%. Can you find a way to fix this ?

eacraft13: Yes, you're completely right. This does not work as intended. I'm going to close this pr. Perhaps there is a way to accomplish creating an x only for the length of the content, as well as setting the img width to a percentage of the item's full available space, but I'm not sure how to as of now.

PR #969, opened 2017-07-01
Only take up width of content (with display: inline-block)

eacraft13:
This allows eliminator x to only span width of content

[tao-4441](https://oat-sa.atlassian.net/browse/TAO-4441)

@ssipasseuth I was hoping you could look at this again. Essentially, I still want to get the 'x' to only cover the area of the content. 

So, the difficulties are still with an img and how a user can set the img width in percentage. When the img is wrapped in a div within .qti-block, then the parent tag .label-content having display: inline-block affects the img width. That is, a width of 25%, for example, is applied twice. Thus, the width of 100% on img tags that are contained in a div. 

There is still one issue, but I personally think it is better than not having this change. When an img tag is directly in .qti-block, .label-content width will still take up the whole area (i.e. a width of 100%). So, in these cases, the eliminator 'x' will still cover 100% of the area. I added another item zip to the jira ticket to help testing.


ssipasseuth:
Agreed, it is better this way. Thanks for the detailed description !

PR #971, opened 2017-07-04
Feature/tao 4554 math support in gapmatch

no-chris:
https://oat-sa.atlassian.net/browse/TAO-4554

How to test:
- create a gapMatch interaction
- add a choice or use the default one
- by clicking on the ckEditor toolbar, you should now be able to add a Math Element into the gap choice.

Alternative choice creation method:
- in the item body, insert a Math expression
- select the math expression
- click the magic wand on the right
- the new choice should contain the math expression


no-chris:
Please hold on on the PR as I need to integrate some remarks from Sam

ssipasseuth:
To mitigate the issue with the toolbar, what about setting a min-width of 80px to the `<li>` of the gap choice ?

no-chris:
I fixed the QTI issue and improved the styling following Sam suggestion.

Most UX issues are referenced in this TT that we created while working on this task: https://oat-sa.atlassian.net/browse/TT-181. Hopefully we'll have time to fix them during this sprint.

Another issue is that the magic wand does not popup when a widget is selected (math, image). That also explains why the image is not taken into account in your video. Unfortunately that would require more in depth modifications to the QtiCreator to implement right now.

no-chris:
Please hold the PR as Sam raised another issue related to the QTI SDK. We'll probably wait for Jérôme on monday to review the QTI SDK PR.

views/js/qtiCreator/editor/ckEditor/htmlEditor.js
```diff
@@ -233,7 +233,12 @@ define([
                     if(typeof options.qtiMedia !== 'undefined'){
                         ckConfig.qtiMedia = options.qtiMedia;
                     }
-
+                    if(typeof options.qtiImage !== 'undefined'){
```
eacraft13: Are we strictly looking for `undefined`? At quick glance, this looks like all this can be replaced with a deep merge.

eacraft13:
Just fyi...
I did a check today and I'm still getting qti error when adding content to an emptied choice. One other issue is that choices with images seem to consume 100% width (I think this issue is worth a look, but it could go into the TT issue as well).
Other than that, I haven't any complaints about the code. Just let me know when it's ready to go.

no-chris:
QTI issue should be fixed with a quick workaround: a placeholder text is now inserted in empty choices. A better way would be to implement validators in the QTI model, but this is out of the scope for now.

PR #976, opened 2017-07-10
Update to QTI-SDK legacy 0.11.0 through lib-tao-qti.

bugalot:
Here is a list of things that might be worth to test:

- After extension/composer update, previously created gapMatchInteraction still work.
- After extension/composer update, previously compiled delivery containing gapMatchInteractions still work.
- No other side effects...

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -56,7 +56,7 @@
     },
     "minimum-stability" : "dev",
     "require": {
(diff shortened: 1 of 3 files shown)


no-chris:
will be included in https://github.com/oat-sa/extension-tao-itemqti/pull/971

PR #986, opened 2017-07-25
Feature/tao 4696 remove shared libs

ssipasseuth:
Requires https://github.com/oat-sa/tao-core/pull/1421

1. Removed shared libraries registry and all dependencies.
2. Upgraded the shared libraries into portable "safe" libraries and reorganized the portable libs to Separate OAT implementation from vendors'.
3. Allow the portable element unit test to keep using the source code.
4. Fixed the rendering of PIC (to allow testing the related PR https://github.com/oat-sa/extension-tao-itemqti-pic/pull/46)


ssipasseuth:
Just found a bug that I fixed with the 2 latest commits. A duplicate for legacyPortableSharedLib is required as some shared libs have dependencies with others and the naming is not compatible with the new one based on `taoQtiItem/portableLib`. Without this fix, existing PCIs from taoDepp for instance will no longer work.

scripts/install/RegisterLegacyPortableLibraries.php
```diff
@@ -0,0 +1,60 @@
...
+
+/**
+ * Register former portable shared libraries to make existing PCI implementation compatible
+ */
+class RegisterLegacyPortableLibraries  extends InstallAction
+{
+    public function __invoke($params)
+    {
+        //register location of portable libs to legacy share lib aliases for backward compatibility
+        $portableSafeLibPath = ROOT_URL.'taoQtiItem/views/js/legacyPortableSharedLib';
```
krampstudio: Can we still use `ROOT_URL` ?

views/js/portableElementRegistry/provider/localManifestProvider.js
```diff
@@ -49,6 +49,24 @@ define(['lodash', 'context', 'core/promise'], function(_, context, Promise){
...
     /**
+     * Use portable element source if exists
+     *
+     * @param {Object} manifest - the manifest of the pci to be modified
+     * @returns {Object} the modified manifest
+     */
+    function useSource(manifest){
+        if(manifest.runtime && _.isArray(manifest.runtime.src)){
+            delete manifest.runtime.hook;//hook is going to be removed with the support of IMS PCI v1
+            manifest.runtime.libraries = manifest.runtime.src;
```
krampstudio: is 'libraries` the replacement of `src` in PCI v1 ?

ssipasseuth: src is basically the uncompiled code from PCI. It is is a TAO internal format. The "libraries" however represents the actual used code for PCIs. FYI, it is TAO specific implementation of PCI. The official PCI uses the modules instead. But it has been agreed that we will keep both implementation working side by side. We will keep improving our implementation and pushing new changes to IMS while supporting the official PCI standard.

scripts/install/RegisterLegacyPortableLibraries.php
ssipasseuth: Hmm, I am checking its usage now.

ssipasseuth: So I replaced the call of $clientLibRegistry->set() to $clientLibRegistry->register() to ensure only relative path is written in the client lib registry config file. Therefore the ROOT_URL usage is safe here as the ROOT_URL will now be replaced with the relative URL.
Example: 
this `ROOT_URL.'taoQtiItem/views/js/legacyPortableSharedLib/jquery_2_1_1'` becomes 
```
'IMSGlobal/jquery_2_1_1' => array(
            'extId' => 'taoQtiItem',
            'path' => 'js/legacyPortableSharedLib/jquery_2_1_1'
        ),
```

ssipasseuth:
Merged develop to that branch and fixed the conflicts. Ready to be merged

scripts/install/RegisterLegacyPortableLibraries.php
```diff
@@ -0,0 +1,60 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2014 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
krampstudio: wrong header date

scripts/install/RegisterLegacyPortableLibraries.php
krampstudio: No, my question was more about the usage of the constant itself. I think there was a task at some point to replace this constant. (please check with a backend)

views/build/grunt/portableelement.js
```diff
@@ -0,0 +1,219 @@
...
+        var model;
+        portableModels.forEach(function(portableModel){
+            if(file.match(new RegExp('\/' + portableModel.file + '$'))){
+                model = {};
+                model.type = portableModel.type;
+                model.manifest = grunt.file.readJSON(file);
+                model.basePath = file.replace('/' + portableModel.file, '');
+                model.id = model.manifest.typeIdentifier;
+                model.runtimeHook = getHookFileName(model.manifest.runtime, model.id);
+                model.minRuntimeFile = getMinHookFile(model.manifest.runtime);
```
krampstudio: You might need to break the loop (ie `return false`)

views/build/grunt/portableelement.js
```diff
@@ -0,0 +1,219 @@
...
+
+        compileTasks = manifests.map(function(file){
+            return new Promise(function (resolve, reject) {
+                var report = [];
+                var config;
+                var model = getPortableModelFromFile(file);
+
+                if(!model){
+                    //not the targeted one
+                    return resolve([grunt.log.error.bind(null, 'invalid portable manifest file ' + file)]);
```
krampstudio: Why do you use `.bind` on all log methods ? Are the methods expecting to run in the lexical scope where `this = grunt` ?

ssipasseuth: It is for the messages to stay grouped and ordered while each async optimization action run.
With a direct call, the messages order are messed up so we can get one message from one compilation task under another one...

views/build/grunt/portableelement.js
ssipasseuth: ok, done

ssipasseuth:
All comments to this PR and its associated ones are processed. Please check again.

scripts/install/RegisterLegacyPortableLibraries.php
ssipasseuth: No it is ok. I have checked with the backend. No plan for removing it so far.

PR #988, opened 2017-07-26
Feature/tao 4594 tables authoring

no-chris:
Related PR:
- https://github.com/oat-sa/ckeditor-dev/pull/6
- https://github.com/oat-sa/tao-core/pull/1424

Please see comments on the JIRA tickets for limitations and known issues
https://oat-sa.atlassian.net/browse/TAO-4594


views/js/qtiCreator/widgets/Widget.js
```diff
@@ -153,6 +153,9 @@ define([
                 enteredStates,
                 i;
 
+            // uncomment to see what's going on
+            // console.log('changing state of ', this.serial, ': ', (currentState || {}).name, ' => ', stateName);
```
ssipasseuth: would it not be better to use the tao logger for this ?

model/qti/ParserFactory.php
```diff
@@ -353,6 +364,20 @@ protected function parseContainerItemBody(DOMElement $data, ContainerItemBody $c
             }
         }
 
+        // parse for tables, but only the ones containing interactions
```
no-chris: That's because tables may contains interactions : the current block will call parseContainerInteractive so the interactions that the tables may contain are properly parsed.

model/qti/ParserFactory.php
```diff
@@ -1385,6 +1410,15 @@ private function buildImg(DOMElement $data){
         return $returnValue;
     }
 
+    private function buildTable(DOMElement $data){
+
+        $attributes = $this->extractAttributes($data);
+        $table = new Table($attributes);
+        $this->parseContainerStatic($data, $table->getBody());
```
no-chris: Tables are parsed in 2 differents places : first we only parse tables that contains interactions, then we parse the other tables (see previous comment)

PR #1015, opened 2017-09-11
Feature/tao 4827 docking ckeditor

ssipasseuth:
Requires https://github.com/oat-sa/tao-core/pull/1462

Docked the html editor toolbar and fixed various toolbar related issue in authoring.
The html editor now are initialized during the question state in creator widgets.


no-chris:
I did a functional review only for now. Works well apart for a few styling issues, some might not be related to this PR, in that case we'll create bugs for those.

Shall we apply the new action style here ? (at least the color, not the border)
actionspropform
resourceselector

The text "add choice" was previously blue, so are text actions in the properties form, is this change wanted or is it a regression?
addchoice

This happens in response mode only: icons disappears once they are applied to a selection
extended text

Action sizing problems:
image
math
shared stimulus

Regressions in Add Choice styling, and also maybe have the actions bordered ?
inline choice

Modal feedback popup does not appear in the right place if the item is scrolled
modalfeedback

This is probably not related to this PR but I just noticed it...
orderliststyle

views/js/controller/creator/index.js
```diff
@@ -47,7 +47,7 @@ define([
             'menu':              $('.menu', $container),
             'menuLeft':          $('.menu-left', $container),
             'menuRight':         $('.menu-right', $container),
-            'title':             $('#item-editor-panel > h1', $container),
+            'title':             $('#item-editor-panel .toolbar-container h1', $container),
```
no-chris: I'm not sure here: why is the title in a toolbar container and the toolbar in the title area ?
The toolbar should be in a 'toolbar' area and the title in a 'title' area. 

Sure we may want to hide the title area when the toolbar is displayed but this is a different concern that the area broker should not be aware of.

views/js/qtiCreator/tpl/toolbars/htmlEditorTrigger.tpl
```diff
@@ -1,5 +1 @@
-<div class="mini-tlb" data-edit="{{state}}" data-for="{{serial}}" data-role="cke-launcher-tlb">
-    <div class="rgt tlb-button" data-role="cke-launcher" title="Formatting options">
-        <span class="icon-font"></span>
-    </div>
-</div>
\ No newline at end of file
+<div class="mini-tlb" data-edit="{{state}}" data-for="{{serial}}" data-role="cke-launcher-tlb"></div>
```
no-chris: is this file still needed?

views/scss/item-creator.scss
```diff
@@ -3,6 +3,9 @@
 
 //colors used accross the item creator
 $itemCreatorSideBarBgColor: mix(#fff, $grey, 85%);
+$itemCreatorSideBarHeaderColor: mix(#fff, $websiteBorder, 70%);
+$itemCreatorWidgetElementBorderColor: mix(#fff, $info, 20%);
+$itemCreatorWidgetElementBackgroundColor: mix(#fff, $info, 90%);
```
no-chris: Could you please clarify what are sideBar and Element ?
If those colors are used for styling the mini toolbar then it might be better to duplicate them in different variable with explicit names:
`$lightBlue: mix(#fff, $info, 90%);`
`$itemCreatorWidgetElementBg: $lightBlue;`
`$itemCreatorActionBtnBg: $lightBlue;`

The fact that they use the same color does not mean that these elements will have their color changed at the same time and for the same new colors.

views/js/qtiCreator/widgets/states/Active.js
```diff
@@ -15,6 +15,7 @@ define([
             if (
                 container !== e.target
                 && !$.contains(container, e.target)
+                && !$.contains($('#item-editor-panel .toolbar-container').get(0), e.target)
                 && !$.contains($('#modalFeedbacks')[0], e.target) //if click triggered inside the #modalFeedback then state must not be changed.
```
no-chris: Should the area broker be used here ?

views/templates/QtiCreator/index.tpl
```diff
@@ -35,13 +34,20 @@ use oat\tao\helpers\Template;
     <!-- item panel -->
     <main id="item-editor-panel" class="clearfix">
 
-        <h1 class="truncate"></h1>
+        <div class="toolbar-container">
+            <h1 class="truncate"></h1>
+            <div id="toolbar-top"></div>
+            <div id="toolbar-bottom"></div>
+        </div>
```
no-chris: is toolbar-bottom used somewhere?

ssipasseuth:
Most issues has been addressed : 
specific color variables are defined to make sass more manageable.
many toolbar styles has been fixed too

The remaining issues are the scrolling issue about modal feedback and the list style for order interaction -> the bugs have been created in JIRA

PR #1039, opened 2017-10-05
Feature/tao 5234/include meta data excel export

ionutpad:


config/default/simpleExporter.conf.php
```diff
@@ -68,5 +72,19 @@
...
         ),
+        'metadataProperties' => array(
+            'extractor' => 'MetaDataOntologyExtractor',
+            'parameters' => array(
+                'valuesAsColumns' => true,
+                'excludedProperties' => array(
+                    taoItems_models_classes_ItemsService::PROPERTY_ITEM_CONTENT,
+                    taoItems_models_classes_ItemsService::PROPERTY_ITEM_MODEL,
+                    taoTests_models_classes_TestsService::PROPERTY_TEST_TESTMODEL,
+                    taoTests_models_classes_TestsService::TEST_TESTCONTENT_PROP,
```
antoinerobin: taoTests isn't required by taoQtiItem (it is the other way, Tests requires Items)
This exclude property should be in the taoQtiTest metadata exporter

manifest.php
```diff
@@ -28,7 +28,7 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '10.5.2',
+    'version'     => '10.5.3',
```
antoinerobin: it is more a feature than a fix

model/Export/Extractor/OntologyExtractorRunner.php
```diff
@@ -0,0 +1,49 @@
...
+{
+    /**
+     * @param core_kernel_classes_Resource $item
+     * @param core_kernel_classes_Property $classProperty
+     *
+     * @return HashEntry
+     */
+    public static function run(core_kernel_classes_Resource $item, core_kernel_classes_Property $classProperty)
+    {
+        $itemProperty = $item->getProperty($classProperty->getUri());
```
antoinerobin: this line doesn't make sense, you are getting the exact same property

scripts/update/Updater.php
```diff
@@ -441,6 +446,34 @@ public function update($initialVersion){
...
+            $options['extractors']['MetaDataOntologyExtractor'] = new MetaDataOntologyExtractor();
+            $options['columns']['metadataProperties'] = [
+                'extractor' => 'MetaDataOntologyExtractor',
+                'parameters' => array(
+                    'valuesAsColumns' => true,
+                    'excludedProperties' => array(
+                        taoItems_models_classes_ItemsService::PROPERTY_ITEM_CONTENT,
+                        taoItems_models_classes_ItemsService::PROPERTY_ITEM_MODEL,
+                        taoTests_models_classes_TestsService::PROPERTY_TEST_TESTMODEL,
+                        taoTests_models_classes_TestsService::TEST_TESTCONTENT_PROP,
```
antoinerobin: these properties shouldn't be there (TEST_MODEL and TESTCONTENT)

model/Export/Extractor/OntologyExtractorRunner.php
```diff
@@ -0,0 +1,49 @@
...
+        $labelItemProperty = $itemProperty->getLabel();
+
+        $extractor = new OntologyExtractor();
+        $extractor->addColumn($labelItemProperty, ['property' => $classProperty->getUri()]);
```
antoinerobin: could you please modify the OntologyExtractor so it takes the uri instead of the label.
example : uriProperty1=>labelA, uriProperty2=>labelA, uriProperty3=>labelA you should see 3 columns with the same labelA, for now you will have only 1 column with the last call

ionutpad: Did not supported, so I modified to add a suffix (1)..(n) if the column with the same name exists

config/default/simpleExporter.conf.php
ionutpad: DONE

manifest.php
ionutpad: Right, 1.6.0 corrected

model/Export/Extractor/OntologyExtractorRunner.php
ionutpad: Here I am trying to get match of a property into a specific item, and then get the label of that value.
The class property are coming from a collection of  properties that this item can have.

PR #1044, opened 2017-10-06
Feature/tao 5146 renamed related item

no-chris:
As discussed during the implementation of https://oat-sa.atlassian.net/browse/TAO-5146

Unit test are passing and no regression has been observed, though more testing might be needed.

This also impacts https://github.com/oat-sa/extension-tao-mp/blob/7e5a53e5efff2026082f8055ecfcbd80e1b50ee5/views/js/pciCreator/dev/adaptiveChoiceInteraction/creator/widget/Widget.js#L51
We should discuss the appropriate strategy between updating this PCI and the taoQtiItem extension on their instance, or locking the taoQtiItem dependency to a version inferior to this one.


ssipasseuth:
Or we could leave this method in place but flag it as deprecated with a warning in the logger and replace as much as possible with the new method as you did for this PR and other in the PIC extension. I believe it would be safer this way.

views/js/qtiItem/core/Element.js
```diff
@@ -233,23 +239,37 @@ define([
...
+                    composingElts[i].setRootElement(item);
                 }
             }
 
         },
-        getRelatedItem : function(){
+        /**
+         * @deprecated - use getRootElement() instead
+         */
+        getReleatedItem : function(){
```
ssipasseuth: oops typo getRelatedItem()

no-chris: nice catch !

PR #1067, opened 2017-11-03
Tao 5491/qti xmllang attribute data not transfered in markup

IlyaYarkavetsOnePoint:
Modified templates to ensure that xml:lang is outputted as html lang attribute in templates;


ssipasseuth:
Ok, the code looks good.
But a quick question, maybe for @bugalot the QTI expert: I am wondering whether in practice, we might set different language in different node or we would just jet it to the root node assessmentItem and use the `xml:lang` there to base the other information upon. I can see some use cases where we want to set different languages within an item (a language item ?) but don't know if it is good practice.
Also that would me that the TTS feature should read the language on each node and dynamically change it, wow.

IlyaYarkavetsOnePoint:
@ssipasseuth , If i understand you: this would be done some time later. At the moment we are just adding support on the item level. Lower levels would be "supported" later as there must be some "rework" for renderers - some item parts, such as prompt, are just stripping out lang attribute. As it was discussed with Mark and Jerome through slack - at the moment we are just adding support on the item level (btw, some work on node levels was done - templates now will output attribute if it exists, but, as I said, backend renderers must be "looked through" to find such problems as attribute stripping), other levels will be added later.
p.s.
Language hierarchy that was discussed:
1) html DOM-element attribute (interface language, as I remember);
2) item level (.qti-item);
3) item-element level;
4) language set in TTS config (as a fallback, if nothing found for current lang);

ssipasseuth:
ok, thanks for the clarificaiton

bugalot:
Hello @IlyaYarkavetsOnePoint I was about to review but I realized there is a conflict in `views/js/qtiCommonRenderer/tpl/tooltip.tpl`. I don't want to break anything so could you please resolve the conflict and contact me when it's done? Thanks in advance !

bugalot:
Oops I did a mistake, I closed the PR by clicking a not appropriate button. I reopned immediately.

PR #1068, opened 2017-11-03
Tao 5505/tts cannot click to speak choice content

IlyaYarkavetsOnePoint:
Now, in choice interactions, if we have tts tool, and click-to-speak function is activated - click-to-speak function will work on clicking the choice option instead of selecting it.


jsconan:
Also, please pay attention to how you assign the reviewers. You should at least select a PR assignee, the PR reviewer is optional. If I did not take a look into the review panel, I'd probably ignore your PR.

IlyaYarkavetsOnePoint:
> Could you please try to use a more generic approach, without targeting the TTS elements?
Or maybe by working directly inside the TTS plugin.
>
Yep, I thought about this moment. But situation is like we have two 'not-so-good' choices:
1) To check if tts plugin exists (`$().length > 0`) and click-to-speak function enabled in config (another `$().length > 0`) and click-to-speak function activated by end user (`$().hasClass('active')`). If all are true - then we are preventing our handler for click to run further;
2) Inside tts plugin on click-to-speak function activation by end user we must find all "choices" (`$('.qti-choice')`) and add some class/attribute to identify that before click-to-speak function will be disabled by test taker we must prevent our handler to run before `stopPropagation` and `preventDefault` will be called inside it. 
Then, inside our handler, we must check for this class/attribute existence and stop further execution if it is present, before `preventDefault/stopPropagation` are called.

IMO, the first way looks a little bit better. First of all, it won't break anything if there will be some js error in, for example, textHelp libs. And yes, I agree, some kind of "coupling" exists in this case - but it's just DOM-element/class existence check.
In second case, there are also not so many changes - just some lines of js code. But... Again, IMO, "coupling" will be more tight - because we need to edit two extensions (this, itemqti, and testqti extension as textHelp lib lays there) and, in future, if there will be some modifications to code we will need to edit code not in one place, but in two separate - and this are to separate extensions. But, if you insist, - I'll do that. Also, if you have some thoughts on more elegant variant of resolution for second case - you are welcome!

>Be aware also that the issue applies for all kinds of choices interactions, not only the simpleChoice.

Yes, that's why not the `simpleChoice` or other `%type%Choice` renderer but common `ChoiceInteraction` renderer was edited, because, as I see, this is the place where common click event handler for all choice types lay.

IlyaYarkavetsOnePoint:
> Also, please pay attention to how you assign the reviewers. You should at least select a PR assignee, the PR reviewer is optional. If I did not take a look into the review panel, I'd probably ignore your PR.
Thank you for the info. I'll draw attention to this in future.

jsconan:
Ok, I understand your point. However it is a bad practice to introduce external specificities in a component. How would you manage that if tomorrow we need to take care of another particularity from another component? This way we would quickly go into contraption like software, with all the drawbracks you can imagine: bad maintainability, bug prone, etc.

A better approach would be to inject feature, by taking the control from the outside. This way, the target component does not have to know about weird specific usage, but only act as it should do.

Your second approach is way better, as it only introduce a new API: a flag that informs the component it should disable a behavior when activated. Obviously, this flag should not introduce any coupling, but should be generic enough (say `preventClick` for instance).

However, this fix needs to recompile every delivery to take care of it. And it should then be better to do no touch the code of the choice interactions, but only take the control from the outside. But yes, this approach is much more complicated. Maybe could you ask @ssipasseuth about that.

ssipasseuth:
Hi added clarification to the bug ticket in JIRA:

> Little clarification: the bug related to all type of choices, in all type of interactions. I took only choice interaction as an example. We need to to have TTS to work on all item contents : be it a simple text block, an image, a math or anything in interaction prompt and choices. e.g. associable choice in associate/match interaction is likely to have the same issue, and maybe PCI contents too. The solution should be generic enough to address all this cases (without code duplication). 
Actually I don't even know if that could even be possible for PCIs, but that probably needs more investigation.

IlyaYarkavetsOnePoint:
Ok, so, as I understand, changes will be next:
1) inside our TTS lib loader, in click-to-speak handler, all interaction elements will be looped through and `prevent-click-handler` class will be added to them.
2) inside interaction renderers there will be just some block like that:
`
if ($_self.hasClass('prevent-click-handler')) {
    return;
}
`
where `$_self` is the reference to the interaction element.
So, there's just one problem left in this case - stop click-to-speak function every time test-taker clicks previous/next/skip buttons - as for now it does not stop click-to-speak from working, and either all elements should be "looped through" again in some kind of callback after item load, or we should disable click-to-speak execution on every item/section load.
I have some thoughts how it should be done for second case (disabling of cts function), so it's just not a big problem, but needs a little research.
Also, there would be another PR in another repo (testqti) for this issue - as I stated above, there would be some fixes for our TTS lib loader on click-to-speak activation.

views/js/qtiCommonRenderer/renderers/interactions/InlineChoiceInteraction.js
```diff
@@ -77,7 +77,17 @@ define([
 
         _setInstructions(interaction);
 
-        $container.on('change', function(){
+        $container.on('change', function(e){
+            //if tts component is loaded and click-to-speak function is activated - we must fix the situation when select2 prevents tts from working
+            //for this a "one-moment" handler of option click is added and removed after event fired
+            if ($(e.currentTarget).parents('.qti-item').hasClass('prevent-click-handler')) {
+                var $selectedIndex = $(e.currentTarget)[0].options.selectedIndex ? $(e.currentTarget)[0].options.selectedIndex : null;
+                $container.find('option').on('click', function(e) {
```
jsconan: Even if you remove the listener just after it should be better to namespace it, to avoid collision with other possible listeners. Or you could use the one-shot listener mode with `.one()`

views/js/qtiCommonRenderer/renderers/interactions/AssociateInteraction.js
```diff
@@ -381,24 +381,46 @@ define([
             // Point & click handlers
 
             interact($container.selector).on('tap', function() {
+                //if tts component is loaded and click-to-speak function is activated - we should prevent this listener to go further
+                if ($(e.currentTarget).parents('.qti-item').hasClass('prevent-click-handler')) {
```
jsconan: Does the element have several parents with the class `.qti-item` ? Otherwise it should be better to use `.closest()` method.

IlyaYarkavetsOnePoint: It is a question not to discuss, but make it clearer for me:
Do we have only one item at a time? I mean that do item with two/three/more interactions inside have only one `.qti-item` container? I'll test this situation a little bit later, if yes - this code will be left as is at the moment, if no - will be changed to `closest`

views/js/qtiCommonRenderer/renderers/interactions/InlineChoiceInteraction.js
IlyaYarkavetsOnePoint: Will update code.

views/js/qtiCommonRenderer/renderers/interactions/AssociateInteraction.js
jsconan: Yes, our implementation of the TestRunner only displays one item at a time.
An item may contains more that one interaction, but the item container is the same for all. Each interaction have its own sub container `.qti-interaction`.

IlyaYarkavetsOnePoint:
Made requested changes.

jsconan:
Works as expected. GG!

PR #1071, opened 2017-11-13
Feature/items bulk import

jsconan:
Add a CLI tool that imports items packages from directory.
Useful to import easily bunch of items without wasting time on the UI, like for CAT items.

### DIFF
--- a/scripts/cli/importItems.php
+++ b/scripts/cli/importItems.php
@@ -1,0 +1,301 @@
+<?php
+/**
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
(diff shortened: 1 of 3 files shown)


scripts/cli/importItems.php
```diff
@@ -0,0 +1,301 @@
...
+ *
+ * @package oat\taoQtiItem\scripts\cli
+ * @author Jean-Sébastien Conan <jean-sebastien@taotesting.com>
+ */
+class importItems implements Action, ServiceLocatorAwareInterface
+{
+    use OntologyAwareTrait;
+    use ServiceLocatorAwareTrait;
+    
+    const ITEM_CLASS_URI = 'http://www.tao.lu/Ontologies/TAOItem.rdf#Item';
```
antoinerobin: we already have the taoOntology::ITEM_CLASS_URI

scripts/cli/importItems.php
```diff
@@ -0,0 +1,301 @@
...
+    protected $processed = 0;
+
+    /**
+     * @param array $params
+     * @return Report
+     * @throws \common_Exception
+     */
+    public function __invoke($params = [])
+    {
+        \common_ext_ExtensionsManager::singleton()->getExtensionById('taoQtiItem');
```
antoinerobin: do you really need it ?

jsconan: It does not seem indeed. I removed it

scripts/cli/importItems.php
jsconan: I was aware of that. Fixed now

jsconan:
Changes made.

PR #1073, opened 2017-11-16
Fix/tao 5546/fix import of file

ionutpad:


bugalot:
@ionutpad I'm not sure to understand what it exactly fixes. Could you describe a little bit more? Maybe with a package I could try?

PR #1075, opened 2017-11-20
Feature/tao 5614/strip namespaces for elements

ionutpad:


model/qti/ImportService.php
```diff
@@ -112,12 +112,11 @@ public function importQTIFile($qtiFile, core_kernel_classes_Class $itemClass, $v
      *
      * @param core_kernel_classes_Class $itemClass
      * @param Item $qtiModel
-     * @param Resource $qtiItemResource
      * @throws common_exception_Error
      * @throws \common_Exception
      * @return core_kernel_classes_Resource
      */
-    protected function createRdfItem(core_kernel_classes_Class $itemClass, Item $qtiModel, Resource $qtiItemResource)
+    protected function createRdfItem(core_kernel_classes_Class $itemClass, Item $qtiModel)
```
jsconan: Is this change related to the feature ?
I feel uncomfortable with that as I have no idea of the possible impacts...

model/qti/container/Container.php
```diff
@@ -366,4 +372,46 @@ public function toArray($filterVariableContent = false, &$filtered = array()){
...
+     * @param $html
+     * @param array $namespaces
+     * @return null|string|string[]
+     */
+    protected function stripNamespaces($html, $namespaces)
+    {
+        $regexArray = [];
+
+        foreach ($namespaces as $namespace) {
+            $regexArray[] = "/[$namespace]+:([$namespace]+[=>])/";
```
jsconan: I am not sure about the resulting regex. I have the feeling it is bugprone.

For instance, if the `$namespace` is `"ht5"`, as in your sample, we will get this regex: `/[ht5]+:([ht5]+[=>])/`
As per the `preg_replace` below, you will replace the matched string by the first expression, that is the namespace itself (and only string like `"ht5:ht5>"` will be mathed).
Moreover, your regex will match any strings like: `"hhhh:555="` or `"55hhtthh:tt>"` , but not `ht5:ruby>`.

A better regex would be: `'/' . preg_quote($namespace) . ':([^=>]+?[=>])/'`

And there is also no protection against possible unauthorized chars.

model/qti/ImportService.php
ionutpad: This is fixing and issue when you try to import a single xml file you get 500 error message.
I have other PR to fix this, but I have to merge that branch into this one to be able to test it.

model/qti/container/Container.php
ionutpad: I did test this regex with <qt5:ruby>test</qt5:ruby> and it was working fine.
I will give a try with your example also, and let you know

ionutpad: Well did a couple of tests more, striping the namespace here it's problematic because this method it's also called when we import a file, So we are going to have the xml imported with namespace stripped. I think we will have to strip namespaces in on the frontend side, much safer. Can you offer some guidance where I can do this?

jsconan: I am not an expert in this area. Maybe @ssipasseuth could help you more than me.
However, for the regex I can help you as I have a good experience in this domain.

PR #1076, opened 2017-11-21
Tao 5547/specify language for tts in items

IlyaYarkavetsOnePoint:
Language for the item can now be specified in item properties at authoring moment;
it will work after pr for TAO-5491 [(pr)](https://github.com/oat-sa/extension-tao-itemqti/pull/1067) be merged


views/js/qtiCreator/helper/itemLoader.js
```diff
@@ -65,6 +65,11 @@ define([
                             loadedItem.setNamespaces(namespaces);
                             loadedItem.setSchemaLocations(qtiSchemaLocation);
 
+                            //add languages list to the item
+                            if (data.languagesList) {
+                                loadedItem.attr('languagesList', data.languagesList);
```
ssipasseuth: please do not use qtiElement.attr() to store data that is not a standard QTI element attribute. For this, you should rather use qtiElement.data() which is used to store metadata associated to the element which does not affect its semantic

views/js/qtiCreator/helper/itemLoader.js
```diff
@@ -86,6 +91,11 @@ define([
                         //tag the item as a new one
                         newItem.data('new', true);
 
+                        //add languages list to the item
+                        if (data.languagesList) {
+                            newItem.attr('languagesList', data.languagesList);
```
ssipasseuth: same as above, please use `newItem .data()` instead.

manifest.php
```diff
@@ -28,7 +28,7 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '11.1.0',
+    'version'     => '11.1.1',
```
ssipasseuth: it is more a feature addition, no ? so follow the semantic versionning, should it rather not be 11.2.0 instead ?

views/js/qtiCreator/helper/itemLoader.js
IlyaYarkavetsOnePoint: Ok, but it leads to another problem: `item.data` is not accessible to template it only has access to `item.attributes` object. I'll try to find a way to pass data/read it from there.

views/js/qtiCreator/helper/itemLoader.js
IlyaYarkavetsOnePoint: Same as above. I'll keep you updated.

IlyaYarkavetsOnePoint:
Made requested changes.
Link to pr added

scripts/update/Updater.php
```diff
@@ -473,7 +473,7 @@ public function update($initialVersion){
             $this->setVersion('10.7.0');
         }
 
-        $this->skip('10.7.0', '11.1.0');
+        $this->skip('10.7.0', '11.1.1');
```
ssipasseuth: don't forget about the updater ;)

IlyaYarkavetsOnePoint: Oh, my fault :). Changed

views/js/qtiCreator/tpl/forms/item.tpl
```diff
@@ -35,3 +35,20 @@
...
+        </label>
+        <span class="icon-help tooltipstered" data-tooltip="~ .tooltip-content:first" data-tooltip-theme="info"></span>
+        <span class="tooltip-content">
+            {{__ "Define item language."}}
+        </span>
+        <select name="xml:lang">
+            {{#each languagesList}}
+                <option value="{{@key}}"{{#equal @key ../xml:lang}} selected="selected"{{/equal}}>{{this}}</option>
+            {{/each}}
+        </select>
```
ssipasseuth: sorry missed that point during the first review but you need to use the select2 here to make it visually consistent with the other properties like in https://github.com/oat-sa/extension-tao-itemqti/blob/3455ccc0304dd4485686c7870762f5307c340765/views/js/qtiCreator/tpl/forms/interactions/textEntry.tpl#L23-L27

IlyaYarkavetsOnePoint: Changed. Missed that point that inside backend we use select2 lib too, not only during delivery

ssipasseuth:
Ok now awaiting the required PR before testing it again.

ssipasseuth:
Please let me know when the other required PR is merged so I can check it again.

ssipasseuth:
Before leaving for holiday, @bugalot @IlyaYarkavetsOnePoint I am assigning the PR to you as you are reviewing the related PR. The code is ok, would just need testing with the other PR.
Sorry for the trouble.

PR #1078, opened 2017-11-22
Feature/tao 5614/strip namespaces in frontend

ionutpad:


views/js/qtiItem/core/Loader.js
```diff
@@ -414,6 +414,10 @@ define([
         },
         loadPicData : function(pic, data){
             loadPortableCustomElementData(pic, data);
+        },
+
+        stripNamespaces: function (body) {
```
jsconan: Sorry, you may not be aware of that, but the function should be named. I know this is a legacy code, and even if the rest of the file is not following the good practices we are following now, could you at least apply it on this function? The reason why is while in debug session, it is difficult to know in which function we are if it is anonymous.

```
stripNamespaces: function stripNamespaces(body) {
...
}
```

views/js/qtiItem/core/Loader.js
```diff
@@ -414,6 +414,10 @@ define([
...
+           return body.replace(new RegExp('(".*?")|(?:[^\s\/<]+?:)', 'g'), '');
```
jsconan: Hmmm. I see an issue here, as the value in quotes will be removed as well as the namespaces. You should put a `$1` in the replace value. The regex contains one named expression `(".*?")` and one anonymous expression `(?:[^\s\/<]+?:)`  (because of the leading `(?:`). So the first expression will provide a parameter, while the second will be ignored. And to keep the value matched by the first expression you should provided the substitution `$1`.

BTW, instead of creating verbosly a regex, you could use the condensed form: `/(".*?")|(?:[^\s\/<]+?:)/g`, so the code should become:

```javascript
stripNamespaces: function stripNamespaces(body) {
    return body.replace(/(".*?")|(?:[^\s\/<]+?:)/g, '$1');
}
```

views/js/qtiItem/core/Loader.js
ionutpad: Sounds good

views/js/qtiItem/core/Loader.js
ionutpad: I truly believe you. Not an expert in regex.
Tested looks stripped.

jsconan:
Works well! Very nice!

PR #1080, opened 2017-11-23
Revert "Feature/tao 5614/strip namespaces in frontend"

ionutpad:
Reverts oat-sa/extension-tao-itemqti#1078

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -28,7 +28,7 @@ return array(
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '11.1.2',
+    'version'     => '11.2.1',
     'author'      => 'Open Assessment Technologies',
     'requires' => array(
         'taoItems' => '>=4.2.4',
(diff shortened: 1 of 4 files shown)


no-chris:
You might want to check the version numbers ;-)

manifest.php
```diff
@@ -28,7 +28,7 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '11.2.0',
+    'version'     => '11.1.2',
```
jsconan: You should instead increase the version

model/qti/ImportService.php
```diff
@@ -117,7 +117,7 @@ public function importQTIFile($qtiFile, core_kernel_classes_Class $itemClass, $v
      * @throws \common_Exception
      * @return core_kernel_classes_Resource
      */
-    protected function createRdfItem(core_kernel_classes_Class $itemClass, Item $qtiModel)
+    protected function createRdfItem(core_kernel_classes_Class $itemClass, Item $qtiModel, Resource $qtiItemResource)
```
jsconan: Does this fix need to be reverted too ?

PR #1081, opened 2017-11-24
Fix/tao 5646/escape for items attributes

zagovorichev:


views/js/qtiXmlRenderer/renderers/Img.js
```diff
@@ -1,6 +1,16 @@
...
-        template : tpl
+        template : tpl,
+        getData : function(item, data) {
+
+            data.attributes = _.mapValues(data.attributes, function (val) {
+                return _.isString(val) ? _.escape(val) : val;
+            });
+
+            return data;
+        }
```
ssipasseuth: Have you tested if the attributes from other QTI elements on top of items and images are impacted ?
What about creating a helper to systematically apply this string protection ?
Moreoever, you are using `_.escape` which sounds reasonable but I saw in a separate PR the use of `core/encoder/entity`, so I am wondering which one is more suitable. Your thought on this ?

zagovorichev: 1. If I understood you correctly then yes, I tried different images in different places, but will be good if you check your thoughts
2. I don't see the reason, why do we need helper for one row code
3. Have no idea, there  was _.escape so I just applied this for each of the attributes

ssipasseuth: I meant different QTI elements having string property values e.g. placeholder for text entry interaction for example.
The helper would be needed only if this problem is more widespread that a couple of places. I Not needed if you con confirm that there is no much more places (other attributes of other QTI elements) with this issue.

ssipasseuth:
We are running out of time so I am merging this for now as it help fixing the currently identified issues.
(The remaining places where such issue may occur are TBD)

PR #1083, opened 2017-12-06
Handle XML namespaces

dietertaotesting:
You can use the item  views/js/test/samples/qtiv2p2/xml_namespaces/with-ns.xml from the unit tests to test this.
Note that said item contains prefixed elements.
Load the item in the editor as well as in the preview. In both cases the ruby elements should have the prefix stripped and be displayed with browser defaults. 
When you save the item nmaespaces and prefixes should be restored


views/js/test/qtiItem/helper/xmlNsHandler/test.js
```diff
@@ -95,11 +95,18 @@ define([
...
+            xml.documentElement.getAttribute('xmlns'),
+            'http://www.imsglobal.org/xsd/imsqti_v2p1',
+            'Initially the QTI version is 2.1'
+        );
 
         // Add some code that requires namespacing. In real life this would for instance come from CK-Editor
         noNsXml = noNsXml.replace('__PLACEHOLDER__', '<ruby><rb>北海道</rb><rt>ほっかいどう</rt></ruby>');
         noNsXml = xmlNsHandler.restoreNs(noNsXml, {});
 
+        console.log(noNsXml);
```
jsconan: is it intentional? :)

PR #1090, opened 2017-12-12
Tao 5727/stateless result storage

boajer:
depends on https://github.com/oat-sa/extension-tao-outcome/pull/101


jbout:
controller no longer in use

PR #1100, opened 2018-01-03
Feature/tao 492 mediasizer

dietertaotesting:
Depends on https://github.com/oat-sa/tao-core/pull/1602
Instructions:
In the item editor create a media interaction and add a video (http://techslides.com/sample-webm-ogg-and-mp4-video-files-for-html5 is a good source). In the right sidebar you should have a slider to resize the video.

There is a separate bug that prevents videos from shrinking below a certain size. You can open a TAO instance without the media sizer to confirm this. Also note that in the QTI.xml the values are correct.


manifest.php
```diff
@@ -28,7 +28,7 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '12.3.0',
+    'version'     => '12.4.0',
```
jsconan: Since there is a dependency on a feature brought by taoCore, you should also update the required version of `tao`

dietertaotesting: resolved

PR #1109, opened 2018-01-30
Fix/tao 5505/tts  cannot click to speak choice content

uncleempty:
Fix for bug with click-to-speak and associate interactions (drag&drop was enabled but not released on clicking at answer when click-to-speak was activated)

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/AssociateInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/AssociateInteraction.js
@@ -380,10 +380,15 @@ define([
 
             // Point & click handlers
 
-            interact($container.selector).on('tap', function() {
+            interact($container.selector).on('tap', function(e) {
                 //if tts component is loaded and click-to-speak function is activated - we should prevent this listener to go further
                 if ($(e.currentTarget).closest('.qti-item').hasClass('prevent-click-handler')) {
+                    isDragAndDropEnabled = false;
                     return;
+                } else {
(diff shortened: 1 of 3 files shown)


views/js/qtiCommonRenderer/renderers/interactions/AssociateInteraction.js
```diff
@@ -380,10 +380,15 @@ define([
...
-            interact($container.selector).on('tap', function() {
+            interact($container.selector).on('tap', function(e) {
                 //if tts component is loaded and click-to-speak function is activated - we should prevent this listener to go further
                 if ($(e.currentTarget).closest('.qti-item').hasClass('prevent-click-handler')) {
+                    isDragAndDropEnabled = false;
                     return;
+                } else {
+                    if (self.getOption && self.getOption("enableDragAndDrop") && self.getOption("enableDragAndDrop").associate) {
+                        isDragAndDropEnabled = self.getOption("enableDragAndDrop").associate;
+                    }
```
ssipasseuth: It would be better to avoid repeating the same piece of code multiple time whenever possible by wrapping this into a function like `self.isDragAndDropEnabled()`. Could you add it into a function that will make the code easier to read. Also you can save one code block level by directly assignation like `else { isDragAndDropEnabled  = self.isDragAndDropEnabled()}`

ssipasseuth (commented):
it works but can you try to make the code simpler and more readable ? this piece of code is getting bigger and bigger.

PR #1112, opened 2018-02-06
Fix/tao 5878 sideloading portable element

ssipasseuth:
Works with https://github.com/oat-sa/extension-tao-testqti/pull/1212

Allow directly load portable element manifests on client side and remove the need for an extra XHR to load PCI data.
The detailed changes are:
* portable element data is compiled during delivery creation
* portable element registration now also copies files into an alias version along with the exact version, e.g. registering v4.1.0, v4.1.1, v4.1.24 will updated the v4.1.*
* the update script with re-register all portable elements
* extra option on the qti item provider allow loading portable element directly (without the need for the portable element providers to make an extra XHR)
* `taoQtiItem/portableElementRegistry/provider/sideLoadingProviderFactory` has been added to support the previous point
* added `PortableElementService:: getPortableElementByClass()` to extra portable element data from a QTI item object

### Functional testing scenarios performed:

**old test runner**
there should be no difference in behaviour before and after the new code update (old & new delivery should work)

**new test runner**
1. from develop, create a test containing PCI & PIC, compile the delivery
2. switch to new codes, run tao update, create a new delivery without deleting the old, both still work
3. upgrade a PCI to a new patch version (e.g. likert v0.4.1 (develop) -> v0.4.2 (see jira ticket attachment)), check that both new and old deliveries manages to get the update
4. upgrade a PCI to a new feature version (e.g. likert v0.4.2 (develop) -> v0.5.0 (see jira ticket attachment)), check that only the old delivery manages to get it

Jira ticket: https://oat-sa.atlassian.net/browse/TAO-5878


krampstudio:
I'll review the functionnal side, but I'd like a backend developer to perform a code review

model/portableElement/PortableElementService.php
```diff
@@ -38,6 +38,9 @@ class PortableElementService implements ServiceLocatorAwareInterface
 {
     use ServiceLocatorAwareTrait;
 
+    const PORTABLE_CLASS_INTERACTION = 'oat\\taoQtiItem\\model\\qti\\interaction\\CustomInteraction';
```
krampstudio: Shouldn't you use the static member `::class` on the real class, once imported  ? 
```php
use oat\taoQtiItem\model\qti\interaction\CustomInteraction;
//...
CustomInteraction::class
```

(Needs validation from a backend dev)

model/portableElement/storage/PortableElementRegistry.php
```diff
@@ -363,11 +369,20 @@ protected function getRuntime(PortableElementObject $object)
...
-     * @return array
-     * @throws PortableElementInconsistencyModelException
+     * Get the alias version for a given version number, e.g. 2.1.5 becomes 2.1.*
+     * @param $versionString
+     * @return mixed
      */
-    public function getLatestRuntimes()
-    {
+    private function getAliasVersion($versionString){
+        return preg_replace('/^([0-9]+\.[0-9]+\.)([0-9]+)$/', '$1*', $versionString);
```
krampstudio: It's not a show-stopper, but handling version with a library would be safer (or do we force registration of `x.x.x` version format ?)
If we don't have one already in tao, there's https://github.com/naneau/semver

krampstudio:
Except the previous comments the code looks good.
The test behavior works as expected, but we should be clear that the originial bug will be fixed only after a delivery compilation, otherwise the call to get the PCI data is still running (which is good for backward compat).
The upgrade of PCIs works as expected, with the new versions, only semver's bugfixes are published. Well done!

model/portableElement/PortableElementService.php
ssipasseuth: you right, it is prettier this way -> done

ssipasseuth:
Ok, so I fixed the unit test by properly clear registered samples on tear down, used the php class name which makes it cleaner indeed and introduced the `naneau/semver` lib to manage version number for PCI more properly.

bugalot:
Seems good to me thanks !

PR #1115, opened 2018-02-15
Feature/tao 5729 send assets to cloudfront

antoinerobin:


jbout:
Introduces a dependency on oat-sa/lib-generis-aws that might not be used by 99% of the users

PR #1116, opened 2018-02-16
`maxWord` constrain now respected on `paste`

dietertaotesting:
Please read the comment regarding CKE in https://oat-sa.atlassian.net/browse/TAO-5946. You can also find a test item in the ticket.

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
@@ -90,7 +90,7 @@ define([
                         editor.config = ckConfigurator.getConfig(editor, toolbarType, ckOptions);
 
                         if(limiter.enabled){
-                            limiter.listenKeyPress();
+                            limiter.listenTextInput();
                         }
                     });
                     editor.on('change', function(e) {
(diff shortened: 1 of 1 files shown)


dietertaotesting:
I moved the relevant to tao/utils and added a test, see https://github.com/oat-sa/extension-tao-itemqti/pull/1116

views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
```diff
@@ -753,6 +823,7 @@ define([
         enable : enable,
         disable : disable,
         clearText : clearText,
-        setText : setText
+        setText : setText,
+        foo: function(){console.log(1239778674696435493)}
```
krampstudio: I think this as been left unintentionally

manifest.php
```diff
@@ -33,7 +33,7 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '13.0.0',
+    'version'     => '13.0.1',
```
krampstudio: You should also now increase the dependency to `tao-core`, but since it's a huge jump you should also increase the version too: 
```
   'version' => '13.1.0'
   'requires' => array(
      //...
        'tao' => '>=17.4.0'
   )
```

PR #1123, opened 2018-02-27
Fix/tao 6044 keep media playing state

krampstudio:
This PR save and restore the media player state within the media interaction. The following values are saved : 
 - the position within the media, 
 - the volume
 - the mute status

Regarding the unit tests the mediaplayer isn't working on PhantomJS so the unit tests falls back in that case (as the mediaplayer unit test does...), but I've kept the test to at least run them in browsers.

To test that particular PR (alone), create a test with non linear test part that contains items with media interactions. Then navigate and play the medias.  
Without the PR the media restarts at it's beginning but with the PR the media restart where you left it.

Please note this PR brings only the state management.

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/MediaInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/MediaInteraction.js
@@ -283,18 +283,15 @@ define([
         var restorePlayerState = function restorePlayerState(playerState){
             if(playerState && interaction.mediaElement){
 
-                //Muted state
-                if(typeof playerState.muted !== 'undefined'){
-                    if(playerState.muted === true){
-                        interaction.mediaElement.mute();
-                    } else {
-                        interaction.mediaElement.unmute();
-                    }
-                }
-
                 //Volume
                 if(_.isNumber(playerState.volume)){
                     interaction.mediaElement.setVolume(playerState.volume);
+                }
+
+                //Muted state (always after the volume)
+                if(_.isBoolean(playerState.muted)){
+                    interaction.mediaElement.mute(playerState.muted);
+                    interaction.mediaElement.startMuted = playerState.muted;
                 }
 
                 //Position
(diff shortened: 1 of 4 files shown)


no-chris:
@jsconan will merge the PR as I'm not there next week.

krampstudio:
@jsconan I've fixed the restore of the mute state

views/js/qtiCommonRenderer/renderers/interactions/MediaInteraction.js
```diff
@@ -314,14 +311,15 @@ define([
...
-                if(interaction.mediaElement.is('ready')){
-                    restorePlayerState(state.player);
-                } else {
+                //if(interaction.mediaElement.is('ready')){
+                    //restorePlayerState(state.player);
+                //} else {
                     interaction.mediaElement.on('ready.state', function(){
                         interaction.mediaElement.off('ready.state');
                         restorePlayerState(state.player);
+                        console.log('READY CALLED');
```
jsconan: Is it wanted, or is it a forgotten debug line?

views/js/qtiCommonRenderer/renderers/interactions/MediaInteraction.js
```diff
@@ -314,14 +311,15 @@ define([
             }
 
             if (_.isPlainObject(state.player) && interaction.mediaElement) {
-                if(interaction.mediaElement.is('ready')){
-                    restorePlayerState(state.player);
-                } else {
+                //if(interaction.mediaElement.is('ready')){
+                    //restorePlayerState(state.player);
+                //} else {
```
jsconan: Forgotten debug change?

jsconan:
I tested the feature with video and audio items. The restore state works well on page reload, not on move next/previous. Is it correct?

krampstudio:
Fixed. 
This PR alone should restore the state only on next/previous
If the answer cache is enabled, it should restore it on both

PR #1124, opened 2018-03-02
Fix/tao 5829/rest publisher role

zagovorichev:
depends on https://github.com/oat-sa/tao-core/pull/1666


zagovorichev:
@ionutpad Update

PR #1125, opened 2018-03-05
Tao 6032/unneeded item language versions clearing

uncleempty:
Added script to deal with unneeded language versions;


jbout (changes requested):
Bypassing the abstractions for performance is not an option

PR #1131, opened 2018-03-19
check for empty string has been fixed

SergiiTao:


model/qti/metadata/ontology/GenericLomOntologyClassificationExtractor.php
```diff
@@ -70,7 +73,10 @@ public function extract($resource)
             $property = $this->getResource($triple->predicate);
             $value = $triple->object;
 
-            if (! empty($value) && $property->isProperty() && ! in_array($property->getUri(), self::$excludedProperties)) {
+            if (trim($value) != ''
```
chilly86: Can $value be an array? If it can be, (what I don't think) then the trim will log an error. Otherwise it works well. :)

PR #1133, opened 2018-03-20
Fix/tao 6072 order interaction cardinality

ssipasseuth:
Per QTI standard: http://www.imsglobal.org/question/qtiv2p1/imsqti_infov2p1.html#element10283
> The orderInteraction must be bound to a response variable with a baseType of identifier and **ordered** cardinality only.

This bug has been introduced by https://github.com/oat-sa/extension-tao-itemqti/commit/e4a094cf1103b40a262bc676308480035ff2146d
This PR just restores the proper usage of the option `updateCardinality ` to ensure that the cardinality of order interaction is always `ordered`. This PR also removes some other useless radio button / checkbox styling code, that is there due to bad copy/paste.

Side note: the graphic order interaction is not suffering from the same bug as it is using this option properly.

### DIFF
--- a/views/js/qtiCreator/widgets/interactions/orderInteraction/states/Question.js
+++ b/views/js/qtiCreator/widgets/interactions/orderInteraction/states/Question.js
@@ -32,9 +32,10 @@ define([
 
     var OrderInteractionStateQuestion = stateFactory.extend(Question);
 
-    OrderInteractionStateQuestion.prototype.initForm = function(updateCardinality){
+    OrderInteractionStateQuestion.prototype.initForm = function(){
 
-        var _widget = this.widget,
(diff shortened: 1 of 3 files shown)


views/js/qtiCreator/widgets/interactions/orderInteraction/states/Question.js
```diff
@@ -70,22 +71,6 @@ define([
 
         interactionFormElement.syncMaxChoices(_widget);
 
-        //modify the checkbox/radio input appearances
-        _widget.on('attributeChange', function(data){
```
ssipasseuth: there is no checkbox nor radio input for order interaction, this is inherited from a copy/paste from the choice interaction

PR #1134, opened 2018-03-21
Fix/tao 6073 extended text empty response

ssipasseuth:
Fix TAO-6073: the QTI model should not reinterpret the values. If it has been set as empty in the model or from the source xml, the output xml should be the same.
So this fix consist in removing this value interpretation: empty value in correct response declaration is no longer removed.
The unit test has been added to ensure that input xml = output xml.

### DIFF
--- a/model/qti/templates/qti.responseDeclaration.tpl.php
+++ b/model/qti/templates/qti.responseDeclaration.tpl.php
@@ -14,7 +14,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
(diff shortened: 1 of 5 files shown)


model/qti/templates/qti.responseDeclaration.tpl.php
```diff
@@ -4,19 +4,19 @@
...
+ *
  * You should have received a copy of the GNU General Public License
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
- * 
- * Copyright (c) 2013 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
- *               
- * 
+ *
+ * Copyright (c) 2013-2017 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
```
krampstudio: We are in 2018, already ;)

PR #1142, opened 2018-04-06
Fix/tao 6142/creator image dimensions fix

uncleempty:
Images width in creator is now looks like it should be - if image width is set to 10% it will look like 10% of width - not like previously when 10% images looked like 100% of width.

### DIFF
--- a/views/js/qtiCreator/widgets/static/img/Widget.js
+++ b/views/js/qtiCreator/widgets/static/img/Widget.js
@@ -66,8 +66,26 @@ define([
             height: this.element.attr('height')
         });
         if (this.$original[0]) {
-            this.$original[0].setAttribute('width', '100%');
-            this.$original[0].setAttribute('height', '100%');
+
+            //previously it looked like
+            //this.$original[0].setAttribute('width', '100%');
+            //this.$original[0].setAttribute('height', '100%');
+            //it was changed to that state to fix the problem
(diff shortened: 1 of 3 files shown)


views/js/qtiCreator/widgets/static/img/Widget.js
```diff
@@ -66,8 +66,26 @@ define([
...
+            //this.$original[0].setAttribute('height', '100%');
+            //it was changed to that state to fix the problem
+            //when we have images inside item look like 100% width,
+            //and no matter was it changed to lower value or not
+
+            var width = this.$original[0].getAttribute('width');
+            if (typeof width === typeof undefined || width.indexOf('%') === -1) {
+                width = '100%';
+            }
+
```
dietertaotesting: `getAttribute` returns `null|""|string` according to https://developer.mozilla.org/en-US/docs/Web/API/Element/getAttribute, so the value will never be `undefined`. A valid width can be either a `number` or a `string` that ends in _%_. I think you should check for `isNaN(width)` instead, you need to try out the correct condition.

On a side note, when checking for `undefined` you need to keep in mind that `undefined` could be overridden with any value, so you cannot use it like that. Correct would `typeof width === 'undefined'`.

uncleempty: Updated the code. Current condition seems to be working like needed.

PR #1144, opened 2018-04-10
Fix/tao 6187 click zone

dietertaotesting:


views/scss/qti/_all-interactions.scss
```diff
@@ -7,6 +7,9 @@
             }
         }
     }
+    .interaction-cover {
+        display:none;
+    }
```
ssipasseuth: again, if it is creator specific, it is better to have only creator scss modified

views/js/qtiCommonRenderer/tpl/interactions/extendedTextInteraction.tpl
```diff
@@ -52,4 +52,5 @@
             </div>
         {{/if}}
     {{/if}}
+  <ins class="interaction-cover"></ins>
```
ssipasseuth: since this bug is affecting the interaction authoring only, I would suggest not affecting the runtime code/template but only the authoring code. 
2 possibilities: 
1. overriding the template in the creator renderer of the extended text interaction
2. append the element (from a template) in the widget intiialization

I tend to find the first solution cleaner.

dietertaotesting: I now did it without any element on the `::after` of the widget. I think that's even nicer.

dietertaotesting:
Thanks!

PR #1159, opened 2018-06-05
Feature/non service compile

jbout:


model/QtiItemCompiler.php
```diff
@@ -57,25 +57,36 @@ class QtiItemCompiler extends taoItems_models_classes_ItemCompiler implements Se
...
+    protected function internalCompile()
     {
-
+        $item = $this->getResource();
         $publicDirectory = $this->spawnPublicDirectory();
         $privateDirectory = $this->spawnPrivateDirectory();
-        $item = $this->getResource();
 
         $report = new common_report_Report(common_report_Report::TYPE_SUCCESS, __('Published %s', $item->getLabel()));
-        if (!taoItems_models_classes_ItemsService::singleton()->isItemModelDefined($item)) {
```
chilly86: what happened with this check? it should be at least in the compile() no?

model/QtiItemCompiler.php
```diff
@@ -57,25 +57,36 @@ class QtiItemCompiler extends taoItems_models_classes_ItemCompiler implements Se
      * @var string
      */
     const INSTANCE_ITEMRUNNER = 'http://www.tao.lu/Ontologies/TAOItem.rdf#ServiceQtiItemRunner';
-    
+
+    /**
+     * {@inheritDoc}
+     * @see taoItems_models_classes_ItemCompiler::compile()
```
chilly86: the return type should be here, because based on the parents I don't really see what could it be

model/QtiItemCompiler.php
```diff
@@ -57,25 +57,36 @@ class QtiItemCompiler extends taoItems_models_classes_ItemCompiler implements Se
...
+            $report->setData($this->createQtiService($item, $publicDirectory, $privateDirectory));
+        }
+        return $report;
+    }
+
     /**
      * Compile qti item
      *
      * @throws taoItems_models_classes_CompilationFailedException
      * @return tao_models_classes_service_ServiceCall
```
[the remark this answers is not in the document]
jbout: What return type would you expect?

model/QtiItemCompiler.php
jbout: Afaict this check is not reachable

model/QtiItemCompiler.php
jbout: Addressed this in a seperate PR: https://github.com/oat-sa/extension-tao-item/pull/314/files

model/QtiItemCompiler.php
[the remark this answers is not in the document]
chilly86: Report as it returns a Report object and you are utilizing the getType method of it

chilly86:
removing registerFS from QtiItemPackerTest will break tests: line 305, 396

PR #1167, opened 2018-06-20
Fix/tao 6396 basic item stimulus import

siwane:
To be tested with and without taoMediaManager installed .Import the test linked in the ticket and at delivery passation image should appear
- Can try with stimulus in subfolder into import package

### DIFF
--- a/model/qti/asset/handler/StimulusHandler.php
+++ b/model/qti/asset/handler/StimulusHandler.php
@@ -24,6 +24,14 @@ use oat\taoItems\model\media\LocalItemSource;
 use oat\taoQtiItem\model\qti\Element;
 use oat\taoQtiItem\model\qti\Item;
 
+/**
+ * Class StimulusHandler
+ *
+ * The purpose of this handler is to extract xml stimulus from item to copy them
+ * If the included stimulus includes images, they should be encoded to avoid file dependencies
+ *
+ * @package oat\taoQtiItem\model\qti\asset\handler
+ */
 class StimulusHandler implements AssetHandler
 {
     /**
@@ -41,6 +49,8 @@ class StimulusHandler implements AssetHandler
     protected $itemSource;
 
     /**
+     * Applicable to all items that contain <xinclude> tag
+     *
      * @param $relativePath
      * @return bool
      * @throws \common_exception_MissingParameter
@@ -70,8 +80,6 @@ class StimulusHandler implements AssetHandler
      * @param $relativePath
      * @return array|mixed
      * @throws \common_Exception
-     * @throws \common_exception_MissingParameter
-     * @throws \tao_models_classes_FileNotFoundException
      */
     public function handle($absolutePath, $relativePath)
     {
@@ -85,6 +93,8 @@ class StimulusHandler implements AssetHandler
     }
 
     /**
+     * Remove ../ to secure path
+     *
      * @param $path
      * @return string
      */
@@ -98,6 +108,8 @@ class StimulusHandler implements AssetHandler
     }
 
     /**
+     * Walk into stimulus file to transform images from path to base64encoded data:image
+     *
      * @param $absolutePath
      */
     protected function encodeStimulusImages($absolutePath)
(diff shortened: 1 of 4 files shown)


model/qti/asset/handler/StimulusHandler.php
```diff
@@ -0,0 +1,189 @@
...
+    }
+
+    /**
+     * @param $attribute
+     * @throws \common_exception_MissingParameter
+     */
+    private function assertAttribute($attribute)
+    {
+        $attributes = get_object_vars($this);
+        if (!isset($attributes[$attribute])) {
```
SergiiTao: the easiest way just to check `if (!isset($this->$attribute))`

model/qti/asset/handler/StimulusHandler.php
```diff
@@ -0,0 +1,189 @@
...
+    {
+        $xincluded = array();
+        /** @var Element $xincludeElement */
+        foreach ($this->getQtiItem()->getComposingElements('oat\taoQtiItem\model\qti\Xinclude') as $xincludeElement) {
+            $xincluded[] = $xincludeElement->attr('href');
+            \common_Logger::i("Xinclude component found in resource '" .
+                $this->getQtiItem()->getIdentifier() . "' with href '" . $xincludeElement->attr('href') . "'.");
+        }
+
+        if(in_array($relativePath, $xincluded)) {
```
SergiiTao: return `in_array($relativePath, $xincluded)` ?

model/qti/asset/handler/StimulusHandler.php
```diff
@@ -0,0 +1,189 @@
...
+     */
+    protected function encodeStimulusImages($absolutePath)
+    {
+        $dom = new \DOMDocument('1.0', 'UTF-8');
+        $dom->loadXML(file_get_contents($absolutePath));
+        $images = $dom->getElementsByTagName('img');
+
+        for($i=0 ; $i<$images->length ; $i++) {
+            $imageFile = dirname($absolutePath) . DIRECTORY_SEPARATOR . ltrim($images->item($i)->getAttribute('src'), DIRECTORY_SEPARATOR);
+            if (file_exists($imageFile)) {
```
SergiiTao: I would use `is_readable`

model/qti/asset/handler/StimulusHandler.php
```diff
@@ -0,0 +1,189 @@
...
+                $encodedSrc = 'data:image/png;base64,' . base64_encode(file_get_contents($imageFile));
+                $images->item($i)->setAttribute('src', $encodedSrc);
+            }
+        }
+        if (isset($encodedSrc)) {
```
SergiiTao: and here and/or `is_writable`

model/qti/asset/handler/StimulusHandler.php
siwane: fixed

model/qti/asset/handler/StimulusHandler.php
siwane: fixed

model/qti/asset/handler/StimulusHandler.php
siwane: Check done at the beginning of the method to avoid process if it is not writable

SergiiTao: indeed :)

model/qti/asset/handler/StimulusHandler.php
```diff
@@ -0,0 +1,188 @@
...
+     * Remove ../ to secure path
+     *
+     * @param $path
+     * @return string
+     */
+    protected function safePath($path)
+    {
+        $safePath = '';
+        if (dirname($path) !== '.') {
+            $safePath = str_replace('../', '', dirname($path)) . '/';
```
SergiiTao: if str_replace find `../` in the middle of the `$path` it will replace it too

siwane: yes to avoid exploit like `toto/../../../../../other/server/file`. It will created a directory with valid part and put it under item directory

PR #1169, opened 2018-07-03
mediasizer improvement

zagovorichev:
- [ ] https://github.com/oat-sa/tao-core/pull/1763


views/js/qtiCreator/widgets/static/img/states/Active.js
```diff
@@ -111,18 +105,59 @@ define([
...
-            if(img.data('responsive') === undefined){
+            if (typeof img.data('responsive') === 'undefined') {
                 if(img.attr('width') && !/[0-9]+%/.test(img.attr('width'))){
                     img.data('responsive', false);
                 }else{
                     img.data('responsive', true);
                 }
             }
 
+            imageUtil.getSize(widget.$original.attr('src'), function(size){
```
krampstudio: Do you always restart from the source ? 
It could be useful to write a simple function to load the media, for example ; 

```js
function loadMedia(imgQtiElement, $imgNode) {
    //if the imgQtiElement has already a width, a height, a type, and the node is loaded 
    return Promise.resolve({{
                  $node:  $imgNode,
                    type: imgQtiElement.attr('type'),
                    src:  imgQtiElement.attr('src'),
                    width:  imgQtiElement.attr('width'),
                    height: imgQtiElement.attr('height'),
     });
    //ottherwise, use the imageUtil to load the required data
}

Comment (krampstudio) [reply views/js/qtiCreator/widgets/static/img/states/Active.js]: Oh and include the initial `responsive` state to it too.

Comment (krampstudio) [reply views/js/qtiCreator/widgets/static/img/states/Active.js]: And extract the unit too.

File: views/js/qtiCreator/widgets/static/img/states/Active.js [hunk 2]
```diff
@@ -111,18 +105,59 @@ define([
...
+                };
+                var options = {
+                    mediaDimension: {
+                        $container: $mediaResizer,
+                        active: true,
+                        responsive: (typeof img.data('responsive') !== 'undefined') ? !!img.data('responsive') : true
+                    }
+                };
+                mediaEditorComponent($mediaSpan.parents('.qti-prompt'), media, options)
+                    .on('change', function (conf) {
```
krampstudio: Why don't you give back the `media` object with the new values on `change` ? 

```js
.on('change', function(media){
  /* 
where media prints the media object
  { 
      responsive : true,
      type: 'image/png',
        src:  'foo.png',
       unit : '%',
       width: '80',
       height: '100',
      ...
  }
*/
});

**Comment by zagovorichev** [comment]
@krampstudio @ssipasseuth Update

**Comment by ssipasseuth** [comment]
Hi Alexander, it seems to work fine except a couple of issues:
1. After a small picture has been added, it stretches to 100% automatically, and I see the resizer component twice with different values. Expected: 1 instance of resizer component and the picture is the same as it would be if we click on reset (initial state)


2. If I change the picture, the new picture takes the size of the old one even if it is much larger. The ratio is also incorrect. I can also see 3 instances of resizer. Expected: still 1 instance of resizer and the size for the new media is recalculated


3. Last little detail to improve the UX, when I add a large picture, it takes the whole size for a moment (routhly 1 second) before it takes the responsive width. If it is not too complex, could you calculate the size of image outside of the viewport or using other solution to prevent the blinking effect ? Doing that will make the app will feel more stable and inspire more confidence

**Comment by zagovorichev** [comment]
@krampstudio @ssipasseuth Update

**Comment by zagovorichev** [comment]
@krampstudio Fixed

**Comment by krampstudio** [comment]
@dietertaotesting I think this has been made in agreement with the UX ppl (@ssipasseuth )

File: views/js/qtiCreator/widgets/static/img/Widget.js [hunk 3]
```diff
@@ -86,6 +91,10 @@ define([
 
             this.$original[0].setAttribute('width', width);
             this.$original[0].setAttribute('height', height);
+            */
+
+            this.$original[0].setAttribute('width', '100%');
+            this.$original[0].removeAttribute('height');
```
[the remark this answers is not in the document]
zagovorichev: the role of this tweak - to save correct values into the QTI. Without this, it will be height="null" or something like that

views/js/qtiCreator/widgets/static/img/states/Active.js
```diff
@@ -104,65 +111,96 @@ define([
...
+            if(imgQtiElement.attr('width') && !/[0-9]+%/.test(imgQtiElement.attr('width'))){
+                imgQtiElement.data('responsive', false);
+            }else{
+                imgQtiElement.data('responsive', true);
+            }
+        }
+
+        if (
+            typeof imgQtiElement.attr('original-width') !== 'undefined'
+            && typeof imgQtiElement.attr('original-height') !== 'undefined'
```
ssipasseuth: what is the use of this `original-width` and `original-height` now ?

ssipasseuth:
The 2 issues reported in the the previous review (preview image ratio and unit test style) are still there. Any explanation for them ?

zagovorichev:
> The 2 issues reported in the the previous review (preview image ratio and unit test style) are still there. Any explanation for them ?

Sorry @ssipasseuth , I've missed that. Fixed.

PR #1172, opened 2018-07-16
Fix/tao 6416/check for loaded theme

uncleempty:
Depends on oat-sa/tao-core#1777

### DIFF
--- a/views/js/qtiRunner/core/Renderer.js
+++ b/views/js/qtiRunner/core/Renderer.js
@@ -71,12 +71,8 @@ define([
         return valid;
     };
 
-    var _getThemesNamespace = function(){
-        return themesHelper.getActiveNamespace();
-    };
-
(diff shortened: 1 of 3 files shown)


views/js/qtiRunner/core/Renderer.js
```diff
@@ -70,6 +71,10 @@ define([
         return valid;
     };
 
+    var _getThemeData = function(){
```
krampstudio: this wrapper is unnecessary, why don't you call just  `var themeData = themesHelper.getCurrentThemeData();`  ? (and by convention we don't prefix with underscore)

krampstudio (changes requested):

views/js/qtiRunner/core/Renderer.js
uncleempty: Removed

PR #1204, opened 2018-11-21
Feature / tao 7150 tooltip

handleman:
task TAO-7150
original parent PR is : https://github.com/oat-sa/tao-core/pull/1880
this repo is affected only in one unit test file.

the reason of I added `$tooltip.qtip('show');` on each html content chef is that the old version was adding html content (display:none;)  every time on plugin initialisation. 
The new version uses more effective approach - it does not manipulate with DOM until first 'open' calling.

related PR:
https://github.com/oat-sa/popper.js/pull/1
https://github.com/oat-sa/extension-tao-dac-simple/pull/78
https://github.com/oat-sa/tao-core/pull/1880


handleman:
@krampstudio  I've made all that were discussed:
Removed all qtip related references,  both from .js files and css styles.

Main PR is : https://github.com/oat-sa/tao-core/pull/1880

Could you do another round of review?

handleman:
@krampstudio tooltip inside qui item padding issue fixed

handleman:
MathJax message fixed

views/js/qtiCreator/widgets/static/math/states/Active.js
```diff
@@ -357,7 +357,7 @@ define([
         });
 
         if(!$mathField.data('$tooltip')){
-            widgetTooltip = tooltip.error($mathField, $tooltipContent, {
+            widgetTooltip = tooltip.error($mathField, $tooltipContent[0].outerHTML, {
```
krampstudio: Shouldn't it be only : 
`var tooltipContent = __('Currently conversion from MathML to LaTeX is not available. Editing MathML here will have the LaTex code discarded.'),`
and
 `widgetTooltip = tooltip.error($mathField, tooltipContent, {`

I don't understand why there's a wrapping `<span>`

handleman: Checked, you are right, there is no need to have DOM element as a message. Fixed

PR #1205, opened 2018-11-26
Feature/tao 6458 add uri to compiled metadata

krampstudio:
Add a shortcut to the item URI into the compiled metadata in order to have it in static deliveries and the mobile app.

How to test this PR : 
 - compile a delivery and check the content of the created `metadataElements.json` files contains the item URI


tikhanovichA:
After compiling a delivery the metadataElements.json file contains an identifier of an item.

PR #1207, opened 2018-11-28
Feature/tao 6458 scorer get qti variables

krampstudio:
In the mobile app we use the scoring library to compute the item outcomes. 
By default the outcomes were produced in the PCI format, for example 
```js
{ RESPONSE : {  list : { identifier: ['choice_1', 'choice_2'] } } }
```
But to comply with the Result API we need information like cardinality or baseType, so we expose the state variable with the outcome.
This PR add the current state to the `outcome event`
The unit tests can be used to check the feature.

Requires https://github.com/oat-sa/extension-tao-item/pull/336

### DIFF
--- a/views/js/test/scoring/provider/test.js
+++ b/views/js/test/scoring/provider/test.js
@@ -1,3 +1,21 @@
+/*
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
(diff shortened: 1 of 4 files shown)


views/js/test/scoring/provider/test.js
```diff
@@ -187,7 +225,7 @@ define([
                 .on('error', function(err){
                     assert.ok(false, 'Got an error : ' + err);
                 })
-                .on('outcome', function(outcomes){
+                .on('outcome', function(outcomes, state){
```
jsconan: state is not tested, is it ok?

krampstudio: Good catch, I forgot to remove it during development, but finally I've added additional test cases.

PR #1211, opened 2018-12-13
Fix/tao 7047/image editing

zagovorichev:
https://oat-sa.atlassian.net/projects/TAO/issues/TAO-7047

### DIFF
--- a/views/js/qtiCreator/editor/ckEditor/htmlEditor.js
+++ b/views/js/qtiCreator/editor/ckEditor/htmlEditor.js
@@ -438,10 +438,10 @@ define([
 
             listenToWidgetCreation = function(){
                 containerWidget.$container
-                    .off('widgetCreated')
-                    .one('widgetCreated', function(e, widgets){
+                    .on('widgetCreated.' + innerWidget.serial, function(e, widgets){
                         var targetWidget = widgets[innerWidget.serial];
                         if(targetWidget){
+                            containerWidget.$container.off('widgetCreated.' + innerWidget.serial);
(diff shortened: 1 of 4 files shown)


views/js/qtiCreator/editor/ckEditor/htmlEditor.js
```diff
@@ -450,6 +450,9 @@ define([
                                     targetWidget.changeState('active');
                                 }
                             }, 100);
+                        } else {
+                            // we need to go through all the widgets until the item is found
+                            listenToWidgetCreation();
```
krampstudio: Isn't this approach a bit dangerous ? 
https://github.com/oat-sa/extension-tao-itemqti/pull/1211/files#diff-788acb8727c68baa11684b324ddb6028L310 seems to be the only place the event is triggered and this location the only place we listen at the event.
So instead of this manipulation, shouldn't it be replaced by something more safe.
I see 2 options : 
 1. keep listening until you have the correct widget : 
```js
var listenToWidgetCreation = function listenToWidgetCreation(){
    var serial  = innerWidget.serial;
    containerWidget.$container.on('widgetCreated.' + serial, function(e, widgets){
          var targetWidget = widgets[innerWidget.serial];
          if(targetWidget){
                   containerWidget.$container.off('widgetCreated.' + serial);
                  //...
            }
     });
}:
``` 
2. change the way the event is triggered so it is contextualized

zagovorichev: perfect, good catch, thank you

zagovorichev:
@krampstudio Update

PR #1212, opened 2018-12-14
added support of the patterns for the ExtendedTextInteraction

zagovorichev:
https://oat-sa.atlassian.net/browse/TAO-7215


views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
```diff
@@ -423,6 +458,34 @@ define([
                 ];
                 var cke;
 
+                createTooltip($container, 'error', __('This is not a valid answer'), true, true);
+                var patternHandler = function patternHandler(e) {
+                    var isCke = _getFormat(interaction) === 'xhtml';
```
zagovorichev: @krampstudio I'm not sure if we still need this, and how to test it at all.

views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
```diff
@@ -32,10 +32,41 @@ define([
...
 
+    /**
+     * Prepare the feedback tooltip for the text input
+     * @param {jQuery} $input
+     * @param {String} theme
+     * @param {String} message
+     * @param {Boolean} [forceCreation=false]
+     * @param {Boolean} [hidden=false]
+     */
+    var createTooltip = function createTooltip($input, theme, message, forceCreation, hidden){
```
zagovorichev: Maybe it doesn't work with CKEditor

krampstudio:
FYI I prefer to wait https://github.com/oat-sa/extension-tao-itemqti/pull/1204 to get merge before

zagovorichev:
@krampstudio PR updated
Could you explain `It seems also the behavior isn't consistent with the "default pattern" (max length and max words constrains are still patterns)` more detailed, please?

krampstudio:
@zagovorichev I'm talking about the constraints like "max words" or "max length" which are only GUI helpers but act like regexp (btw if you set "max length" and switch to "pattern"  you'll have the pattern for the previous constraint). And I think the tooltip should be also displayed for those situations. No?
image

views/js/qtiCommonRenderer/renderers/interactions/ExtendedTextInteraction.js
krampstudio: I think it's required because when the interaction is using XHTML format, the textarea is replaced by CKEditor and the value can't get retrieved the same way.

PR #1218, opened 2019-01-11
fix/TAO-1764/media-player-stop-when-leaving

start-admin:
https://oat-sa.atlassian.net/browse/TAO-1764

Create an item with Media interaction
Upload audio/video file
Press Play on the file
Click "Manage items" button to go back to Items extension - playback goes on

Audio should not be played after closing


views/js/qtiCreator/plugins/navigation/back.js
```diff
@@ -48,6 +48,7 @@ define([
             var itemCreator = this.getHost();
 
             itemCreator.on('exit', function(){
+                $(".item-editor-drop-area").remove();
```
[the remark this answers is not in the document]
start-admin: But destroy not called on exit.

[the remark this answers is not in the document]
krampstudio: So you can listen the `exit` event in the itemCreator.

views/js/qtiCreator/itemCreator.js
```diff
@@ -193,6 +193,10 @@ define([
                     });
                 });
 
+                this.on('exit', function() {
+                    $(".item-editor-drop-area").remove();
```
krampstudio: May be `$('.item-editor-item').empty()` could be more appropirate. @ssipasseuth  could you please validate ?

ssipasseuth: `.item-editor-item` looks like a more suitable selector as it is higher level than `.item-editor-drop-area`.
Also, wouldn't it be better to scope the selector, by using the area broker for instance ?

krampstudio: agree, @start-admin could you please test with `$('.item-editor-item', areaBroker.getItemPanel()).empty()` (see https://github.com/oat-sa/extension-tao-itemqti/blob/3f373fffcaaf9cc885d44e980a25641e5daf7141/views/js/controller/creator/index.js#L47)

PR #1226, opened 2019-01-28
The warning message after setting 'Allowed choices' fixed

start-admin:
https://oat-sa.atlassian.net/browse/TAO-5073

Steps to reproduce:
1. Log in as admin.
2. Go to the item page, create an item, and go to authoring mode.
3. Add Hotspot Interaction and and choose an image file.
4. Define hotspot areas using the tools on the right. (for example 3 areas).
5. Set '1' for Min and '3' for Max in 'Allowed choices'.
6. Click 'Save'.
7. Click 'Preview'.
Actual result: The interaction is displayed with the message 'You must at least 1 choice'. (please, see attachment)
Expected result: The interaction is displayed with the message 'You must select at least 1 choice'.

The warning message after setting 'Allowed choices' fixed

### DIFF
--- a/views/js/qtiCommonRenderer/helpers/instructions/instructionManager.js
+++ b/views/js/qtiCommonRenderer/helpers/instructions/instructionManager.js
@@ -182,7 +182,7 @@ define([
             }
 
             if(!minInstructionSet && min > 0 && (choiceCount === false || min < choiceCount)){
-                msg = (min <= 1) ? __('You must at least %d choice', min) : __('You must select at least %d choices', min);
+                if (min <= 1) msg = __('You must select at least %d choices', min);
                 self.appendInstruction(interaction, msg, function(){
                     if(getResponse(interaction).length >= min){
                         this.setLevel('success');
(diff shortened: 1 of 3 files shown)


views/js/qtiCommonRenderer/helpers/instructions/instructionManager.js
```diff
@@ -182,7 +182,7 @@ define([
             }
 
             if(!minInstructionSet && min > 0 && (choiceCount === false || min < choiceCount)){
-                msg = (min <= 1) ? __('You must at least %d choice', min) : __('You must select at least %d choices', min);
+                msg = __('You must select at least %d choices', min);
```
dietertaotesting: The message should be depending on the number of min choices
`msg = (min <= 1) ? __('You must select at least %d choice', min) : __('You must select at least %d choices', min);` would be correct

start-admin: of course. this is my mistake.

PR #1229, opened 2019-02-04
Fix/tao 5139/select point interaction score popup closing

start-admin:
https://oat-sa.atlassian.net/browse/TAO-5139


views/js/qtiCreator/widgets/interactions/selectPointInteraction/states/Map.js
```diff
@@ -196,7 +196,14 @@ define([
                 } else {
                     //move the score back the shape and show the popup
                     shape.toFront();
-                    $('#score-popup-' + shape.id).show(); 
+                    $('#score-popup-' + shape.id).show();
+
+                    //the click on the cross hides the popup
+                    $('.mapping-editor').on('click', '.closer', function(){
```
krampstudio: this kind of binding should be scoped, ie `$('.mapping-editor', $container)`

views/js/qtiCreator/widgets/interactions/selectPointInteraction/states/Map.js
```diff
@@ -196,7 +196,14 @@ define([
...
+                        $('#score-popup-' + shape.id).hide();
```
krampstudio: You should keep a ref to the element instead of querying the DOM multiple times : 
```
var $scorePopup = $('#score-popup-' + shape.id)
```

views/js/qtiCreator/widgets/interactions/selectPointInteraction/states/Map.js
```diff
@@ -196,7 +196,14 @@ define([
...
+                    });
+
+                    tooltip.lookup($('.panel'));
```
krampstudio: It appears to be not working.
@handleman something forgotten during the tooltip replacement ?

start-admin: This tooltip was not initialized anywhere.
in the import there was an old 'ui / tooltipster' and I replaced it with the actual 'ui / tooltip' and initialized it.
works for me.
maybe he didn't notice.

start-admin:
@krampstudio task is fixed, could you check?

PR #1232, opened 2019-02-08
FIX/TAO-1764/playback continues when switch to Preview

start-admin:
https://oat-sa.atlassian.net/browse/TAO-1764

If start watching video in Authoring mode and then switch to Preview - playback continues.

### DIFF
--- a/views/js/qtiCreator/plugins/menu/preview.js
+++ b/views/js/qtiCreator/plugins/menu/preview.js
@@ -72,7 +72,7 @@ define([
                 text : __('Preview'),
                 cssClass: 'preview-trigger'
             })).on('click', function previewHandler(e){
-                $(document).trigger('open-preview');
+                $(document).trigger('open-preview.qti-item');
 
                 e.preventDefault();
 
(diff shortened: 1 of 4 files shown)


krampstudio:
@martijn-tao could you please take care of the review ? I'll be in holidays

martijn-tao:
> @oatymart could you please take care of the review ? I'll be in holidays

Will do :)

views/js/qtiCreator/plugins/menu/preview.js
```diff
@@ -72,6 +72,8 @@ define([
                 text : __('Preview'),
                 cssClass: 'preview-trigger'
             })).on('click', function previewHandler(e){
+                $(document).trigger('open-preview');
```
oatymart: `$(document).trigger` is not an ideal practice.

However I didn't find a better way to talk to the widget from the preview.

You should at least namespace this event (`open-preview.qti-item`) so that the listener wouldn't be triggered by some unrelated code in the future.

views/js/qtiCreator/widgets/states/Active.js
```diff
@@ -49,6 +49,10 @@ define([
             _widget.changeState('sleep');
         });
 
+        $(document).on('open-preview', function(){
+            _widget.changeState('sleep');
```
oatymart: Same here.

PR #1233, opened 2019-02-12
Fix/tao 7652/rendering issues with tooltip in textreader

ricproenca-tao:
related to https://oat-sa.atlassian.net/browse/TAO-7652
need to backport to sprint95


views/scss/qti/_tooltip.scss
```diff
@@ -25,4 +25,11 @@
 
 .qti-item .tooltip{
     padding: 10px;
+}
+
+.tooltip-active {
+    position: static;
+    border-style: none;
+    padding: 0 0 0 0;
```
jsconan: You could just say: `padding: 0;`

views/scss/qti/_tooltip.scss
```diff
@@ -25,4 +25,11 @@
...
+    font-size: 1em;
```
jsconan: Is the `em` unit accurate here? Unless good reason (for instance size should be relative to parent), we should use `rem` instead.

views/scss/qti/_tooltip.scss
```diff
@@ -25,4 +25,11 @@
 
 .qti-item .tooltip{
     padding: 10px;
+}
+
+.tooltip-active {
+    position: static;
```
jsconan: Caution, this is introducing a side effect, the tooltip will be always placed at the same position in the page, and won't be displayed close to the related target.

image

views/scss/qti/_tooltip.scss
ricproenca-tao: yes

views/scss/qti/_tooltip.scss
ricproenca-tao: The `em` is accurate because in this case we want the to be size of the parent which should be the tetx size.

PR #1242, opened 2019-02-21
Fix/tao 7667/tooltips cover one another rollback n n tooltip

ricproenca-tao:
related to https://oat-sa.atlassian.net/browse/TAO-7667
- rollback  the changed behavior (adding n/n tooltip);
- fix the actual bug.

Unit test: `/taoQtiItem/views/js/test/qtiCommonRenderer/interactions/textEntry/test.html`

### DIFF
--- a/views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js
+++ b/views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js
@@ -36,29 +36,34 @@ define([
     'use strict';
 
     /**
-     * Delete the text input tooltip
+     * Hide the tooltip for the text input
      * @param {jQuery} $input
      */
-    var deleteTooltip = function deleteTooltip($input){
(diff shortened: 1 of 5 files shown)


views/js/qtiCommonRenderer/renderers/interactions/TextEntryInteraction.js
```diff
@@ -226,6 +234,15 @@ define([
     };
 
     var destroy = function destroy(interaction){
+
+        $("input.qti-interaction.qti-inlineInteraction.qti-textEntryInteraction").each(function(index, el) {
```
jsconan: Is it needed to have such a precise selector? Couldn't it wok with only `input.qti-textEntryInteraction`?

ricproenca-tao: Yes, it can!

PR #1249, opened 2019-03-07
hidden non-actions in controllers

zagovorichev:


controller/QtiCreator.php
```diff
@@ -54,7 +51,11 @@
  */
 class QtiCreator extends tao_actions_CommonModule
 {
-    use EventManagerAwareTrait;
+    protected function getEventManager()
```
martijn-tao: Please add a docblock

PR #1257, opened 2019-03-21
Fix/tao 7637 droppable.js placeholder removed from bottom

rkarolis:
https://oat-sa.atlassian.net/browse/TAO-7637

In item authoring, by dragging interaction the placeholder on the item page should be removed by going through bottom.
Minor refactoring.

### DIFF
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -14,7 +14,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2016 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2014-2019 (original work) Open Assessment Technologies SA;
  *
  */
 
@@ -391,6 +391,6 @@ class Updater extends \common_ext_ExtensionUpdater
             $this->setVersion('16.0.0');
         }
 
-        $this->skip('16.0.0', '19.3.0');
+        $this->skip('16.0.0', '19.3.1');
     }
 }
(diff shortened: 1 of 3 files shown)


scripts/update/Updater.php
```diff
@@ -14,7 +14,7 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2016 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2014-2019 (original work) Open Assessment Technologies SA;
```
ferenckiss89: Why 2014? it should be either 2016-2019 if that's the original date or simply 2019 with the latest changes.

views/js/qtiCreator/editor/gridEditor/droppable.js
```diff
@@ -305,8 +305,10 @@ define([
         ).on('mouseleave.gridEdit.gridDragDrop',
             _.debounce(function(e){
                 var goingTo = e.relatedTarget|| e.toElement; //browser compatibility
+                var $target = $(e.target);
 
-                if($(e.target).parents('.qti-itemBody').length) {
+                if($target.parents('.qti-itemBody').length ||
```
ferenckiss89: I'd put also $target && .. just to make sure but if we did not used to, then it's not necessary.

scripts/update/Updater.php
rkarolis: because original code is dating from 2014

views/js/qtiCreator/editor/gridEditor/droppable.js
rkarolis: No we dont do and over here

rkarolis: $target is always here, otherwise there would be no event

PR #1264, opened 2019-04-05
[TAO-7935] WK-1281 PG - TCiaB - highlighter behavior for touchscreen

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-7935
 
Do not discard selection on touchend during highlighting
Allow do clear highlighted node by clicking on it
Change cursor when  highlighting is activated
  
#### How to test
- Run delivery as a test taker on touchscreen device with enabled highlighting tool
- Activate highlighting tool and start selecting the text
- After triggering selection you should be allowed to resize it

- Run delivery as a test taker with enabled highlighting tool
- Highlight some text 
- After click on highlighted text it should be cleared

- Run delivery as a test taker on desktop browser with enabled highlighting tool
- After activating highlighting tool cursor above test item should be changed to marker icon

#### Dependencies
   
Companion PR :
 - [ ] https://github.com/oat-sa/tao-core/pull/2069
 - [ ] https://github.com/oat-sa/extension-tao-testqti/pull/1476


views/scss/qti/_qti-base.scss
```diff
@@ -117,6 +117,10 @@ $mainContainer : '.qti-item';
                 background: $errorBgColor;
             }
         }
+
+        &.highlighter-cursor {
+            cursor: url("../img/qtiIconsPng/marker.png") 0 32, auto;
```
zagovorichev: do you think this is the right way to add icons?
I know only this way: https://hub.taotesting.com/styleguide/icons (don't know if anything changed)

AntonTsymuk: it is not an icon, but static asset

PR #1267, opened 2019-04-12
[WIP] Fix/tao 7637 drag drop item authoring droppable.js

rkarolis:
https://oat-sa.atlassian.net/browse/TAO-7637

Fixing item authoring drag and drop for firefox. It should be able to drag and drop multiple interactions at any position in the item.

Also `:last selector for Safari error fix. Mentioned in TAO-8099


views/js/qtiCreator/editor/gridEditor/droppable.js
```diff
@@ -262,7 +262,7 @@ define([
             }, 50)
         ).on('mousemove.gridEdit.gridDragDrop',
             _.throttle(function(e){ // '[class^="col-"]:not(.new-col), [class*=" col-"]:not(.new-col)
-                var goingTo = e.relatedTarget|| e.toElement; //browser compatibility
+                var goingTo = e.relatedTarget || e.toElement || e.target; //browser compatibility
```
ferenckiss89: Do we really need to check these 3? Would not be enough to have only e.target? If yes, don't we need these in lines: 308 and 241 also?

PR #1271, opened 2019-04-23
Fix/tao 7123 text reader pci render bug

Sunbus:
Task - [TAO-7123](https://oat-sa.atlassian.net/browse/TAO-7123)
This task have important comments, it would be great to check it.

**Current behavior**: when inserting a table within a TextReader PCI, the table does not render correctlyAlso, the contents in the TR column do not seem to update correctly on changes.

**Expected behavior**: tables render as they do outside of the TextReader PCIChanges are updated correctly.

I got that for use `container.render()` we should use container as attribute. 

Checked on core extension.

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -33,7 +33,7 @@ return array(
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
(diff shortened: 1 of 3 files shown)


oatymart:
Closing as the appropriate fix was never found here. Best of luck to the next one who tries!

PR #1273, opened 2019-04-24
fix for Response Processing

zagovorichev:
all `<setOutcomeValue>` elements from the `<responseProcessing>` elements needs to be used in the `<outcomeDeclaration>` elements.

- [x] depends on https://github.com/oat-sa/qti-sdk/pull/135


bugalot:
As soon as this is made `configurable` and not enabled by default, I have no objections!

PR #1283, opened 2019-05-28
Fix/tao 8389/hide qticreator error tooltip

oatymart:
https://oat-sa.atlassian.net/browse/TAO-8389

There were 2 bugs relating to tooltip management in the item authoring.
1. The same error tooltip was being duplicated 1,2,3...10 times or more, preventing it from ever being hidden. This made it seem like negative values were considered invalid (even though they were always valid, and could be saved as such).
2. Error tooltips were remaining on the page after Response Mapping mode was exited.

Both errors are fixed, by:
1. Checking if a form element already has a tooltip and removing it before assigning a new one.
2. Checking on a widget's state change, if it is an interaction leaving the 'map' state. If so, hiding its tooltips.

Background information on QTI widget states: https://hub.taotesting.com/articles/qti/qti-item-creator

To test: go into item authoring and play with changing the response processing mode, and setting invalid (non-numeric) values and negative values in the inputs (negatives are valid!)

### DIFF
--- a/views/js/qtiCreator/widgets/Widget.js
+++ b/views/js/qtiCreator/widgets/Widget.js
@@ -177,14 +177,12 @@ define([
 
             if(currentState){
                 // hide widget tooltips when interaction leaves response mapping ('map') state:
-                if (this.serial.match(/^interaction/)) {
-                    if (currentState.name === 'map' && state.name !== 'map') {
-                        $(self.$container)
-                            .find('[data-has-tooltip]')
-                            .each(function(j, el) {
-                                $(el).data('$tooltip').hide();
-                            });
-                    }
+                if (this.serial.match(/^interaction/) && currentState.name === 'map' && state.name !== 'map') {
+                    this.$container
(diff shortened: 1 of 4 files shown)


views/js/qtiCreator/widgets/Widget.js
```diff
@@ -176,6 +176,16 @@ define([
             }
 
             if(currentState){
+                // hide widget tooltips when interaction leaves response mapping ('map') state:
+                if (this.serial.match(/^interaction/)) {
+                    if (currentState.name === 'map' && state.name !== 'map') {
```
ferenckiss89: Unnecessary nesting of if statements

views/js/qtiCreator/widgets/Widget.js
```diff
@@ -176,6 +176,16 @@ define([
...
+                        $(self.$container)
```
ferenckiss89: ‘this’ can be used here instead of ‘self’

PR #1284, opened 2019-05-31
Feature/nex 81 es2015 migration

krampstudio:
Related to https://oat-sa.atlassian.net/browse/NEX-81

 - Bundles with babel
 - migrate one controller to ES2015

Requires PR : 

 - [x] https://github.com/oat-sa/tao-core/pull/2156

Please note, when trying to run the bundles, if you encounter an issue with MathJax, please just touch the file (it's a very old known issue)


views/build/grunt/bundle.js
```diff
@@ -19,10 +19,13 @@
 /**
```
ferenckiss89: update copyright dates please

views/js/controller/creator/index.js
```diff
@@ -41,8 +41,8 @@ define([
      * Set up the areaBroker mapping from the actual DOM
```
ferenckiss89: update copyright date

views/js/controller/creator/index.js
```diff
@@ -95,14 +95,14 @@ define([
             //load plugins dynamically
             if (config) {
                 if(config.plugins){
-                    _.forEach(config.plugins, function (plugin) {
+                    _.forEach(config.plugins, plugin => {
```
ferenckiss89: lodash could be already removed

views/js/controller/creator/index.js
```diff
@@ -95,14 +95,14 @@ define([
             //load plugins dynamically
             if (config) {
                 if(config.plugins){
-                    _.forEach(config.plugins, function (plugin) {
```
anshul119: Any reason to use the foreach func from lodash instead of the one from array prototype?

views/js/controller/creator/index.js
krampstudio: `plugins` may be an `Object` not an `Array`

views/js/controller/creator/index.js
```diff
@@ -120,15 +120,11 @@ define([
                 //build a new item creator
                 itemCreatorFactory(config, loadAreaBroker(), pluginLoader.getPlugins())
                     .on('error', reportError)
-                    .on('success', function(message){
-                        feedback().success(message);
-                    })
+                    .on('success', message => feedback().success(message) )
                     .on('init', function(){
```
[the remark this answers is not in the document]
krampstudio: actually the lexical scope is used within the function : `this.render` and we don't have it in the closure

views/js/controller/creator/index.js
krampstudio: yes plugins can be an array or an object

PR #1287, opened 2019-06-11
Fix/prevent fe test to hang

krampstudio:
Some tests were failing because they were relying on a fake mock from tao-core which was moved away. 
It appeared a simple mock was sufficient in the target tests. 

Please run the extension test suite  : 
```
cd tao/views/build
npx grunt connect:test taoqtiitemtest
```


views/js/test/qtiCreator/renderers/static/tooltip/test.js
```diff
@@ -38,31 +34,40 @@ define([
...
-                    toolbar: $('#toolbar-top')
-                }
-            }),
-            tooltipContent = 'my tooltip <strong>content</strong>',
-            tooltipSerial = '_tooltip_4568613547893',
-            tooltipId = 'tooltip_123456',
+            getContentCreatorPanelArea() {
+                return $('#item-editor-panel');
+            },
+            getItemPanelArea() {
```
btamas: It returns with the same element like `getContentCreatorPanelArea`, only the selector is different. Is that correct?

krampstudio: yes, I've just took the selectors from the original test

PR #1289, opened 2019-06-14
Added/fixed russian translations

tikhanovichA:


locales/es-MX/messages.po
```diff
@@ -0,0 +1,1829 @@
...
+msgstr " Rellenar los huecos en un texto a partir de un conjunto de opciones."
+
+msgid " For Resource "
+msgstr ""
+
+msgid " or "
+msgstr ""
+
+msgid "%d characters allowed"
+msgstr ""
```
bartlomiejmarszal: WHy this is empty?

tikhanovichA: This is just the copy of existing Spanish translations. We didn't translate all strings, only that we needed for NCCER

PR #1290, opened 2019-06-20
Bugfix/tao 7842 dont allow empty items in delivery

k-v-vladimir:
image

https://oat-sa.atlassian.net/browse/TAO-7842

For Premium version the tests with empty items can't be compiled and published


model/QtiJsonItemCompiler.php
```diff
@@ -86,6 +86,11 @@ protected function deployQtiItem(
         // retrieve the media assets
```
ferenckiss89: Update the date in the license header from 2016 to 2016-2019

ferenckiss89:
> This looks good to me. Unit test for this is not suitable imho. Look at the code:
> 
> * Singleton
> * a dozen of statements in a try
> * some exceptions caught by generic classes
> * writing to filesystem
> * new QtiItemPacker() statement
> * ...
>   It would require a big refactoring, and the functional test example provided by @vladimir Kalskov does the job.
>   It would make sense either to write an integration test maybe, or if we really want to add a unit test upon what Vladimir wrote because we want to leave the code in a better shape than it was before, we could extract @vladimir Kalskov’s lines to a public method and test this method, but it's basically just testing a count()...

We will create a technical debt ticket to do the refactoring and proper testing of this part. It cannot belong to the scope of the bug ticket this time.

ferenckiss89:
Technical debt ticket: https://oat-sa.atlassian.net/browse/TTD-9

PR #1301, opened 2019-06-27
Override metadata properties for each interaction

Alroniks:
Ticket is https://oat-sa.atlassian.net/browse/SD-44
Alternative https://oat-sa.atlassian.net/browse/SI-398

### What does it do?

In case when an item has several interactions it loops through them and applies metadata for each row.
Before it worked only with the first item.


Alroniks:
Please, have a look @bugalot

PR #1303, opened 2019-07-04
[TAO-8590] WK-1253 TCiaB: Line breaks are shown as <br>

AntonTsymuk:
Related to : https://oat-sa.atlassian.net/browse/TAO-8590
 
Since from ckEditor event we get html as value, when paste value use insertHtml method instead of insertText
  
#### How to test
- Prepare a test item with extended text interaction of XHTML format and with some constraints or recommendations
- Add some text with line breaks
- Try to cut and paste the text
- There should be no html tags in the text interaction and text should have line breaks like before it was cut out


handleman:
works well, code looks fine

PR #1304, opened 2019-07-04
Fix/tao 8190/empty item creator reference

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TAO-8190

Fix issue in the `itemCreator`, where the reference of the component wasn't properly returned.

To test:
```
cd tao/views/build
npx grunt connect:test taoqtiitemtest
```

### DIFF
--- a/views/js/qtiCreator/context/qtiCreatorContext.js
+++ b/views/js/qtiCreator/context/qtiCreatorContext.js
@@ -96,6 +96,7 @@ define([
              * @returns {Promise}
              */
             destroy: function destroy() {
+                var self = this;
                 return pluginRun('destroy').then(function(){
 
                     /**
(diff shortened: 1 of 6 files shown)


views/js/qtiCreator/context/qtiCreatorContext.js
```diff
@@ -96,6 +96,7 @@ define([
              * @returns {Promise}
              */
             destroy: function destroy() {
+                var self = this;
```
anshul119: why not use arrow func and then we dont need to store scope anymore

jsconan: Yes, I could have updated the whole file to be ES6. But this is only a fix, and since it took me some time to figure out the issue, I did not want to break something else aside by changing too much things.

PR #1306, opened 2019-07-09
Fix/Transpiler issue with math keyword

jsconan:
Fix an issue occurring with the Math model when transpiling the source code. `Math` is reserved, and using this name for a variable will conflict with global context when transpiling.


views/js/qtiCreator/model/Math.js
```diff
@@ -2,7 +2,7 @@ define([
     'lodash',
```
ferenckiss89: license header is missing here

PR #1308, opened 2019-07-09
Fix version to take care of backport 20.4.3

jsconan:
Due the change in version for the backport #1307, we need to interleave the version bump to allow later update.


ferenckiss89:
Seems good to me

PR #1311, opened 2019-07-12
Feature/tao 8379/encapsulate classes to services

bartlomiejmarszal:
I'm working on service that requires to use these services to provide frontend config


model/service/CreatorConfigService.php
```diff
@@ -0,0 +1,40 @@
...
+use oat\oatbox\service\ConfigurableService;
+use oat\taoQtiItem\model\CreatorConfig;
+
+/**
+ * Creator Config service encapsulation
+ * Class CreatorConfigService
+ * @author Bartlomiej Marszal
+ *
+ */
+class CreatorConfigService extends ConfigurableService
```
Skyld: Please rename it to factory

model/service/HookRegistryService.php
```diff
@@ -0,0 +1,40 @@
...
+use oat\oatbox\service\ConfigurableService;
+use oat\taoQtiItem\model\HookRegistry;
+
+/**
+ * Encapsulating HookRegistry into service
+ * Class HookRegistryService
+ * @author Bartlomiej Marszal
+ *
+ */
+class HookRegistryService extends ConfigurableService
```
Skyld: Please rename it to factory

PR #1334, opened 2019-09-09
Fix TAO-7643 Correct color on hover state

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-7643

Added color on hover state for Choice, Associate, Gap Match, Hottext Interactions.
Added color for tables.

How to reproduce:
- create an item with different interactions
- open Style Editor
- set dark background color and light text color
- save and preview


IgorSalevan:
Looks good.
The only things left to fix are:
- common interactions:
`order`: list bullets get hidden on hover both in right and left list containers, when {color: white; background: black}
- custom interactions:
`calculator`: `math-entry-placeholder` color my conflict with background-color, if color is set to white.

PR #1336, opened 2019-09-17
fix/TAO-9085/minChoices and maxChoices can be equal to the size of choices

bziondik:
Related to: https://oat-sa.atlassian.net/browse/TAO-9085

**Should be merged after PR: https://github.com/oat-sa/tao-item-runner-qti-fe/pull/22**

The issue:  if `maxChoices = size of choices` then condition `if (max > 0 && max < size) {` didn't match and  `} else if (min > 0 && min < size) {
        size = min;
 }` So the size set in minChoices
https://github.com/oat-sa/tao-item-runner-qti-fe/blob/f7d7bb46f3d9e579dd8011215a942454709d07a7/src/qtiCommonRenderer/renderers/interactions/GraphicOrderInteraction.js#L115-L126
https://github.com/oat-sa/extension-tao-itemqti/blob/02b6cd67f00ffe5e2ca9d1ee2403a2e351e94023/views/js/qtiCreator/widgets/interactions/graphicOrderInteraction/Widget.js#L78-L90

Fix:
In calculation, minChoices and maxChoices can be equal to the size of choices.
**Changes in 2 files in `tao-item-runner-qti-fe` and `extension-tao-itemqti`**

How to test:

1. Create an item
2. Drag and Drop a Graphic Order Interaction to the canvas
3. Select the BG
4. Create 2 figures
5. Check the Min and Max choices in "Allowed choices", on the right side, so that Max=1, Min=1
6. In Response mode, choose first figure - figure gets digit 1, second figure - max choices warning
7. In Question mode, modify Max to 2 
8. In Response mode, try to choose one more figure

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -34,7 +34,7 @@ return array(
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '21.3.8',
+    'version'     => '21.3.9',
     'author'      => 'Open Assessment Technologies',
     'requires' => array(
         'taoItems' => '>=9.1.0',
--- a/scripts/update/Updater.php
+++ b/scripts/update/Updater.php
@@ -438,6 +438,6 @@ class Updater extends \common_ext_ExtensionUpdater
             $this->setVersion('21.0.0');
         }
 
-        $this->skip('21.0.0', '21.3.8');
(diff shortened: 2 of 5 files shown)


bziondik:
the version of @oat-sa/tao-item-runner-qti was updated  to 0.4.5

PR #1356, opened 2019-11-18
Fix/uio 1 set minimum value for slider

btamas:
Related to https://oat-sa.atlassian.net/browse/UIO-1, https://oat-sa.atlassian.net/browse/TAO-7302 and https://oat-sa.atlassian.net/browse/TAO-6443

Test:
1. Go to item authoring
2. Set negative or not integer value for slider border values
3. Try to set text as lowerBound or upperBound

### DIFF
--- a/views/js/qtiCreator/tpl/forms/interactions/slider.tpl
+++ b/views/js/qtiCreator/tpl/forms/interactions/slider.tpl
@@ -1,18 +1,18 @@
 
 <div class="panel">
     <label for="lowerBound" class="spinner">{{__ "Lower Bound"}}</label>
-    <input name="lowerBound" value="{{lowerBound}}" data-increment="1" data-min="0" data-validate="$integer" type="text" />
+    <input name="lowerBound" value="{{lowerBound}}" data-increment="1" data-min="0" type="text" />
     <span class="icon-help tooltipstered" data-tooltip="~ .tooltip-content:first" data-tooltip-theme="info"></span>
     <span class="tooltip-content">{{__ "The lower bound of the slider"}}</span>
     <label for="upperBound" class="spinner">{{__ "Upper Bound"}}</label>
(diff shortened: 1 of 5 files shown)


views/js/qtiCreator/tpl/forms/interactions/slider.tpl
```diff
@@ -1,11 +1,11 @@
 
 <div class="panel">
     <label for="lowerBound" class="spinner">{{__ "Lower Bound"}}</label>
-    <input name="lowerBound" value="{{lowerBound}}" data-increment="1" type="text" />
+    <input name="lowerBound" value="{{lowerBound}}" data-increment="1" data-min="0" validator="$integer" type="text" />
     <span class="icon-help tooltipstered" data-tooltip="~ .tooltip-content:first" data-tooltip-theme="info"></span>
     <span class="tooltip-content">{{__ "The lower bound of the slider"}}</span>
     <label for="upperBound" class="spinner">{{__ "Upper Bound"}}</label>
-    <input name="upperBound" value="{{upperBound}}" data-increment="1" type="text" />
+    <input name="upperBound" value="{{upperBound}}" data-increment="1" data-min="0" validator="$integer" type="text" />
```
oatymart: Do you think we should use the integer validator on the step field as well?

Since we only use the integer responseType for this interaction, and any float value you put in as the step is automatically rounded down, I think it would make the UX better to give visual feedback when the user's input is invalid.

btamas: Removed.

views/js/qtiCreator/widgets/interactions/sliderInteraction/states/Question.js
```diff
@@ -25,58 +25,82 @@ define([
         var callbacks = {};
```
ferenckiss89: missing license header

oatymart: Added :wave:

PR #1361, opened 2019-12-02
Added support for APIP 1.0 final

julien-sebire:
This adds support for item type `imsqti_apipitem_xmlv2p2` and manifest validation for APIP 1.0 final (not candidate).

To test, try to import the package attached to this ticket:
https://oat-sa.atlassian.net/browse/TAO-9682
Given the size of the package, you might need to enable the task queue and change your PHP configuration (`upload_max_filesize` and `post_max_size`).


scripts/update/Updater.php
```diff
@@ -438,6 +438,6 @@ public function update($initialVersion){
             $this->setVersion('21.0.0');
         }
 
-        $this->skip('21.0.0', '22.0.7');
+        $this->skip('21.0.0', '22.1.0');
```
zagovorichev: should we update the ontology?

PR #1364, opened 2019-12-10
add getApipData to itemRunner - Acces the data about APIP the content and related media files

lecosson:
Related to: https://oat-sa.atlassian.net/browse/TAO-9637
Acces the data about APIP the content and related media files

update packages versions
#### Dependencies
Requires :
 - [ ] https://github.com/oat-sa/tao-item-runner-qti-fe/pull/34


jsconan (changes requested):
The target link in `package.json` is ok.
But why did you remove the `package-lock.json` file???
It must be versioned.

When working with git branch instead of version, we must only touch the `package.json`.
And the `package-lock.json` file must be updated only once a proper version is available and set in the `package.json`.

PR #1365, opened 2019-12-11
added converter of the apip accessibility data (from xml to array)

zagovorichev:
### Pull request summary
 
*Related to*: https://oat-sa.atlassian.net/browse/TAO-9799
 
Correct compilation of the APIP items with attached files.
 
#### How to test

Create delivery from the test that has APIP items with media files attached. Make sure that JSON has correct data provided and all media files were copied to the compiled delivery.

Requires :
 - [x] https://github.com/oat-sa/tao-core/pull/2353
 - [x] https://github.com/oat-sa/extension-tao-item/pull/384

### DIFF
--- a/model/QtiJsonItemCompiler.php
+++ b/model/QtiJsonItemCompiler.php
@@ -20,14 +20,18 @@
 
 namespace oat\taoQtiItem\model;
 
+use common_exception_Error;
 use common_report_Report;
 use core_kernel_classes_Resource;
+use Exception;
 use oat\taoQtiItem\model\pack\QtiItemPacker;
 use oat\taoQtiItem\model\qti\exception\XIncludeException;
 use oat\taoQtiItem\model\qti\Service;
 use oat\taoQtiItem\model\qti\Element;
+use tao_helpers_Xml;
 use tao_models_classes_service_StorageDirectory;
 use oat\taoQtiItem\model\portableElement\PortableElementService;
+use taoItems_models_classes_CompilationFailedException;
 
 /**
  * The QTI Json Item Compiler
@@ -51,7 +55,8 @@ class QtiJsonItemCompiler extends QtiItemCompiler
 
     /**
      * Generate JSON version of item
-     * @return array consists of item URI, public directory id and private directory id
+     * @return common_report_Report
+     * @throws taoItems_models_classes_CompilationFailedException
      */
     public function compileJson()
     {
@@ -65,7 +70,7 @@ class QtiJsonItemCompiler extends QtiItemCompiler
     }
 
     /**
-     * Desploy all the required files into the provided directories
+     * Deploy all the required files into the provided directories
      *
      * @param core_kernel_classes_Resource $item
      * @param string $language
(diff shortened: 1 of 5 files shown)


model/QtiJsonItemCompiler.php
```diff
@@ -124,13 +130,30 @@ protected function deployQtiItem(
...
+     * Convert internal parameters to json if needed
+     * @param $data
+     * @return array
+     * @throws common_exception_Error
+     */
+    protected function convertXmlAttributes($data) {
+
+        if (is_array($data) && array_key_exists('core', $data)
+            && array_key_exists('apipAccessibility', $data['core'])) {
+
```
yuryfilippovich: according psr 5.1 : 
if (
    $expr1
    && $expr2
) {

yuryfilippovich: if core will not array we will get error: so we can add is_array($data['core']) checking or replace 
 array_key_exists('core', $data) && array_key_exists('apipAccessibility', $data['core']) to isset($data['core']['apipAccessibility'])

model/qti/AssetParser.php
```diff
@@ -105,9 +104,22 @@ public function extract(){
...
         }
+        $this->extractApipAccessibilityAssets();
         return $this->assets;
     }
 
+    private function extractApipAccessibilityAssets()
+    {
+        if (property_exists($this->item, 'apipAccessibility')) {
+            try {
+                $assets = tao_helpers_Xml::extractElements('fileHref', $this->item->getApipAccessibility(), 'http://www.imsglobal.org/xsd/apip/apipv1p0/imsapip_qtiv1p0');
```
yuryfilippovich: psr 2.3

zagovorichev: ?

model/QtiJsonItemCompiler.php
```diff
@@ -124,13 +130,30 @@ protected function deployQtiItem(
...
             return new common_report_Report(
                 common_report_Report::TYPE_ERROR, $e->getMessage()
             );
         }
     }
 
+    /**
+     * Convert internal parameters to json if needed
+     * @param $data
+     * @return array
```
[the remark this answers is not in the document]
zagovorichev: I've changed return statement (as method is protected, I need this check to avoid incorrect behaviour on inheritance)

model/QtiJsonItemCompiler.php
zagovorichev: I added condition that data['core'] is array , I don't want to use isset as it is not checking that key exists, it just checks that result is not null

model/qti/AssetParser.php
zagovorichev: @yuryfilippovich could you explain me what don't you like here and from when TAO supports psr2.3?

yuryfilippovich: limit for one line - 120 characters. look like codeclimate find the same issue

zagovorichev: fixed

zagovorichev:
@yuryfilippovich Update

PR #1373, opened 2020-01-03
Breaking/tdr 9/handlebars 4 upgrade

oatymart:
https://oat-sa.atlassian.net/browse/TR-3264
https://oat-sa.atlassian.net/wiki/spaces/FOUN/pages/144081031/Upgrade+Handlebars

This extension is a bit more complex because it contains the `portableLib` folder for PCIs, and the `portablelement` grunt task definition.

- [ ] bump dependencies in package.json
- [x] rebundle JS
- [ ] composer dependency update


views/js/portableLib/handlebars_4.js
oatymart, line 1: This file is a copy of one of the handlebars 4.7.8 modules installed in `node_modules`. I was in 2 minds, but I placed it here in `portableLib` (beside handlebars 1.3.0) to match the way some current PCIs (not all) are resolving these "portable" modules. If this is ok, I could remove `handlebars.js` (unless you think we should hold onto it for backwards compatibility).

views/js/portableLib/OAT/util/tpl.js
oatymart, line 1: I'm pretty sure this file is never even called by any of the PCIs I've tested.

views/build/moduleWriter.js
```diff
@@ -34,7 +34,7 @@ define([], function () {
     return function moduleWriter(moduleName, compiled) {
         let handlebars = 'handlebars';
         if (isPciRuntime(moduleName)) {
-            handlebars = 'taoQtiItem/portableLib/handlebars';
+            handlebars = 'taoQtiItem/portableLib/handlebars_4';
```
jsconan, line 37: **issue:** The idea looks very good. However, we need to make sure that `taoQtiItem/portableLib/handlebars_4` is loaded in the first place. Without this, the runtime will declare a dependency on an inexistent package...

See https://github.com/oat-sa/extension-tao-itemqti-pci/pull/379#discussion_r1343981613

jsconan (commented):
The idea of using a different namespace for Handlebars in PCI is good. However, it will only work if we can make sure the lib is loaded before. This is the tricky part: we rely on the same mechanism to load the helpers both for the PCI compiler and the regular TAO compiler. The regular compiler will preload the system default Handlebars, we need it for TAO related bundles. When discovering a PCI we need to replace it by the portableLib, but at this stage it is already too late as the RequireJS dependencies tree is already built... If you have some idea, you are welcome :) 

Look at https://oat-sa.atlassian.net/browse/ADF-1293 for having a use case.

views/js/portableLib/OAT/util/tpl.js
jsconan: Indeed, I did not find any usage. However, some 3rd-party PCI providers could use it.

swiec:
@mike-ionut-mihai-sandu-tao just to remind you, that you still didn't pay my invoice.

PR #1399, opened 2020-03-06
Fix tao 7508 match interaction buttons overlap

AnastasiaArcadia:
**Related to:**
https://oat-sa.atlassian.net/browse/TAO-7508

**Description:**
Make Match Interaction buttons not overlap title text field in Item Authoring.

**How to check:**
Create new item with Match Interaction.
Fill in the Text Field.
Text Field should be clearly visible.


jsconan:
This is nice! But could you:
- add the compiled CSS files?
- add a version bump?

PR #1403, opened 2020-03-10
Fix/udi 97/export items user does not gave access

SergiiTao:
depends on oat-sa/tao-core#2428

Ticket: https://oat-sa.atlassian.net/browse/UDI-97

There was a possibility to fetch resources without having permissions to them.
I've created a service to fetch resources safely.


model/Export/ExportForm.php
```diff
@@ -56,108 +57,146 @@ abstract class ExportForm extends tao_helpers_form_FormContainer
...
+        return strtolower(tao_helpers_Display::textCleaner($name, '*'));
+    }
+
+    protected function getOptions(core_kernel_classes_Resource ...$resource): array
+    {
+        $itemService = taoItems_models_classes_ItemsService::singleton();
+        $options = [];
+        $disabledOptions = [];
+        foreach ($resource as $instance) {
+            if ($itemService->hasItemModel($instance, [ItemModel::MODEL_URI])) {
```
wazelin: Early return is preferable here IMO, too much of nesting is happening here
Consider doing `if (!$itemService->hasItemModel($instance, [ItemModel::MODEL_URI])) break;`

SergiiTao: `continue?` or i don't see something?

wazelin: Right, `continue`

model/Export/ApipPackageExportHandler.php
```diff
@@ -57,25 +62,28 @@ public function getLabel()
...
+     *
+     * @return tao_helpers_form_Form
+     * @throws common_Exception
+     * @throws common_exception_Error
      */
     public function getExportForm(core_kernel_classes_Resource $resource)
     {
         if ($resource instanceof core_kernel_classes_Class) {
-            $formData = ['class' => $resource];
+            $formData['items'] = $this->getResourceService()->getAllChildren($resource);
```
wazelin: Not only the form should exclude the items, but the export request handler itself should filter out inaccessible resources.
Right now, I can easily download items, I don't see on that form by calling `/taoItems/ItemExport/index`, extending the list of `instances` manually.

SergiiTao: fixed

wazelin: Right, my bad, misplaced the original comment.
Good job!

PR #1404, opened 2020-03-11
Feature/nec 72/indication file has successfully been uploaded

AndreyNegrov:
___Related to task:___ https://oat-sa.atlassian.net/browse/NEC-72

___Description:___ Indication a file has successfully been uploaded

__Depends on:__ https://github.com/oat-sa/tao-item-runner-qti-fe/pull/56

Translation added


locales/lt-LT/messages.po
```diff
@@ -1575,6 +1575,9 @@ msgstr "Įvertinimo formatas nėra skaitmeninis"
 msgid "The selected file is ready to be sent."
 msgstr "Pasirinktas failas paruoštas išsiųsti."
 
+msgid "Your file has been uploaded."
```
tikhanovichA: What about EN locale? Why I don't see it there?

tikhanovichA: and what about the old string 'The selected file is ready to be sent'? I see that it was translated for all locales

PR #1408, opened 2020-03-17
UDI-26: Update QTI library to a fixed version

wazelin:
UDI-26 fix(dependencies): update `oat-sa/lib-tao-qti`

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -61,7 +61,8 @@
     },
     "minimum-stability": "dev",
     "require": {
-        "oat-sa/lib-tao-qti": "~5.0.0",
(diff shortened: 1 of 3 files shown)


composer.json
```diff
@@ -61,7 +61,8 @@
     },
     "minimum-stability": "dev",
     "require": {
-        "oat-sa/lib-tao-qti": "~5.0.0",
+        "oat-sa/lib-tao-qti": "dev-fix/UDI-26-allow-minor-dependencies-upgrades as 5.0.1",
+        "qtism/qtism": "fix/UDI-26-polygon-intersection as 0.19.1",
```
wazelin: It's here temporarily.
Will update to a stable version of `oat-sa/lib-tao-qti` once released, removing a direct `qtism/qtism` dependency.

PR #1411, opened 2020-03-19
Fix/import shared stimulus

julien-sebire:
This PR fixes two problems in the shared stimulus import:
- Check that the shared stimulus in the package is contained in the uncompressed filesystem path to avoid exposing the whole filesystem to users.
- Actually share the asset imported from one package with several links to it, instead of having one copy for each item using it.

This PR doesn't solve the tree structure issue in the media manager. As to now, the shared stimulus is imported in the media manager in a folder named after the item's name. It should be reproducing the tree structure in the item bank to which the package is imported. This will be a separate PR.

Requires https://github.com/oat-sa/generis/pull/754


model/qti/ImportService.php
```diff
@@ -425,7 +426,7 @@ public function importQtiItem(
         array $metadataInjectors = [],
         array $metadataGuardians = [],
         array $metadataClassLookups = [],
-        array $sharedFiles = [],
+        array &$sharedFiles = [],
```
wazelin: The behavior is getting even more unpredictable here :D

julien-sebire: You're perfectly right. I removed the default value.

PR #1412, opened 2020-03-19
Feature/ms 131/support longinterpretation attribute

anshul119:
Ticket: https://oat-sa.atlassian.net/browse/MS-131

**What has changed?**

* Long interpretation attribute is added
* Validation for the attribute is added (accepts valid URI)
* Tests are added

**How to test?**

* Create an item
* Add an interaction
* Go to response tab
* in the right sidebar, add an outcome declaration
* You should be able to see a new attribute by the name Long Interpretation
* Add a valid URI
* Save the item and export
* The item XML should have the long interpretation attribute present with the value entered by you 

**Run unit tests**

`cd tao/views/build`

`npx grunt connect:test qunit:single --test=/taoQtiItem/views/js/test/qtiCreator/plugins/panel/outcomeEditor/test.html`


views/js/test/qtiCreator/plugins/panel/outcomeEditor/test.js
```diff
@@ -109,8 +109,50 @@ define([
...
+
+        loader.loadItemData(item_airport, function(loadedItem){
+            const itemCreator = creatorMock($container, {}, loadedItem);
+            const pluginInstance = outcomeEditorPlugin(itemCreator, itemCreator.getAreaBroker());
+
+            pluginInstance.init();
+
+            assert.expect(2);
+
+            itemCreator.getAreaBroker().getContainer().on('initResponseForm.outcome-editor', () => {
```
btamas: Can `itemCreator.getAreaBroker().getContainer()` change? I think you can create a value for it in the beginning.

views/js/qtiCreator/widgets/helpers/validators.js
```diff
@@ -69,6 +69,22 @@ define([
...
                 }
             }
+        },
+        //Define a validator that check validity of a URL
+        {
+            name: 'isValidUrl',
+            message: __('Invalid URL'),
+            validate: function validate(value, callback) {
+                if(value) {
+                    try{
```
btamas: I think the `try catch` is not necessary, because you can throw the error as well. Like above.

model/qti/OutcomeDeclaration.php
```diff
@@ -22,7 +22,6 @@
 
 namespace oat\taoQtiItem\model\qti;
 
-use oat\taoQtiItem\model\qti\OutcomeDeclaration;
 use oat\taoQtiItem\model\qti\VariableDeclaration;
 use oat\taoQtiItem\model\qti\Item;
 use oat\taoQtiItem\model\qti\response\Composite;
```
Haru0: All of those use statements are unused. Maybe they could be removed as well (the same thing was done to ` model/qti/attribute/Interpretation.php`)?

views/js/qtiCreator/tpl/outcomeEditor/listing.tpl
```diff
@@ -18,6 +18,16 @@
             <div class="tooltip-content">{{__ "A human interpretation of the variable's value."}}</div>
             <input name="interpretation" value="{{interpretation}}" type="text">
         </div>
+        <div class="panel longinterpretation">
+            <label for="longInterpretation" class="has-icon">{{__ "Long interpretation"}}</label>
+            <span class="icon-help tooltipstered" data-tooltip="~ .tooltip-content:first" data-tooltip-theme="info"></span>
+            <div class="tooltip-content">{{__ "An optional link to an extended interpretation of the outcome variable."}}</div>
+            <input name="longInterpretation"
+                   placeholder="https://www.tao.org/rules.pdf"
```
Haru0: Wouldn't that be better to have more generic url example? Something like `https://example.com/doc.pdf`.
Initially, I thought ist some kind of technical reference. Also, can we do something about such cases as below? 
image

views/js/test/qtiCreator/plugins/panel/outcomeEditor/test.js
```diff
@@ -94,33 +95,76 @@ define([
...
+                    $outcomes.children().last().find('[data-role="edit"]').trigger('click.outcome-editor');
+                });
+
+                itemContainer.find('.adder').trigger('click.outcome-editor');
+            });
+
+            itemContainer.trigger('initResponseForm.outcome-editor');
         });
     });
 });
```
Haru0: A missing blank line at the end of the file.

views/js/qtiCreator/tpl/outcomeEditor/listing.tpl
anshul119: agreed

PR #1420, opened 2020-03-27
Fix/NEC-84/pci instance creation

andrey-ponamarev:
**Related to task:**  https://oat-sa.atlassian.net/browse/NEC-84
**Description:** The problem is all PCI properties with Object type are overrides by empty Object.
**Requires:** https://github.com/oat-sa/extension-tao-itemqti-pci/pull/193
**How to test:** 
- Install `extension-tao-itemqti-pci` version [fix/TAO-10102/liquid-interaction-preview](https://github.com/oat-sa/extension-tao-itemqti-pci/pull/193)
- Create item with liquids interaction
- Open item in preview 
- Click "Submit" button without item interaction
- Expected behavior: There are no errors in console

**Unit tests cli:** `npx grunt connect:test taoqtiitemtest`

### DIFF
--- a/views/js/runtime/qtiCustomInteractionContext.js
+++ b/views/js/runtime/qtiCustomInteractionContext.js
@@ -16,7 +16,7 @@
  * Copyright (c) 2014-2020 (original work) Open Assessment Technlogies SA (under the project TAO-PRODUCT);
  *
  */
-define(function(){
+define(function() {
     'use strict';
 
     //need a global reference to have pciHooks shared in two different requirejs context ("default" and "portableCustomInteraction")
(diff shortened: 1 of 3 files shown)


views/js/runtime/qtiCustomInteractionContext.js
```diff
@@ -75,31 +75,14 @@ define(function(){
          * @param {string} pciTypeIdentifier
          * @returns {Object} clonedPciModel
          */
-        createPciInstance : function(pciTypeIdentifier){
+        createPciInstance: function(pciTypeIdentifier) {
+            var registeredPCI = window._pciHooks[pciTypeIdentifier];
```
btamas: const

views/js/runtime/qtiCustomInteractionContext.js
```diff
@@ -75,31 +75,14 @@ define(function(){
...
-                }
-
-                return instance;
-
-            }else{
+            if (!registeredPCI) {
                 throw new Error('no portable custom interaction hook found with the id ' + pciTypeIdentifier);
             }
+
+            return Object.assign({}, registeredPCI);
```
jsconan: The previous version was somehow doing a clone of nested objects. Should not it be the same here?

views/js/runtime/qtiCustomInteractionContext.js
```diff
@@ -75,31 +75,14 @@ define(function(){
          * @param {string} pciTypeIdentifier
          * @returns {Object} clonedPciModel
          */
-        createPciInstance : function(pciTypeIdentifier){
+        createPciInstance: function(pciTypeIdentifier) {
```
jsconan: You should use object shorthand here. It will fix the code style issue with the anonymous function and will be compliant with ES6 code style as well.

views/js/runtime/qtiCustomInteractionContext.js
```diff
@@ -75,31 +75,14 @@ define(function(){
...
-                        instance[name] = proto[name].constructor();
-                    }else{
-                        //not an object (nor a function) : e.g. 0, 123, '123', null, undefined
-                        instance[name] = proto[name];
-                    }
-                }
-
-                return instance;
-
-            }else{
```
krampstudio: wooh, this looks like we are in the history of JavaScript, like this how our parent were writting code  :rofl:

views/js/runtime/qtiCustomInteractionContext.js
krampstudio: Do you think  `instance[name] = proto[name].constructor();`  was creating a deep clone ?

andrey-ponamarev: There are no problems with that.
The current method `createPciInstance` is factory for creating `registeredPCI` copies.
It simply provides copy of PCI interface implementation.
The issue with `instance[name] = proto[name].constructor();` it's overrides all PCI instance properties which has value `Object` type by empty object.

For instance, I have PCI:
```javascript
const customPCI = {
   init(){
        this.getState()
   },
   getState(){
       return this._state
   },
   _state: {
         value: 'Default state'
   }
}

const newPCI = createPciInstance(customPCI)

newPCI.init();
// Console output 
// without bugfix
// {}
// with bugfix
// { value: 'Default state'}
```

views/js/runtime/qtiCustomInteractionContext.js
andrey-ponamarev: Time is running everything is changed 🤷‍♂️

views/js/runtime/qtiCustomInteractionContext.js
jsconan: I don't think `instance[name] = proto[name].constructor();` was creating a deep clone. However, it could have recreated the object if it was instanciated... (but without the`new` keyword I'm not so sure). Anyway, since it was just creating empty objects, I guess it's ok.

However, I'm still wondering if a simple assign will not introduce issue with object nested inside a PCI's definition. If 2 instances of the same PCI are living in the page, and each of them are mutating this object, what will happen?

andrey-ponamarev: The possible issue which @jsconan mentioned above is not reproducible for liquids PCI since [the internal state assigned to the new value on update](https://github.com/oat-sa/extension-tao-itemqti-pci/blob/master/views/js/pciCreator/dev/liquidsInteraction/runtime/liquidInteraction.js#L291) (state is immutable object).
To prevent issue with mutability in the future, PCI creation is using `deepClone` for now.
Thanks @jsconan for good catch  👍

PR #1422, opened 2020-04-03
Feature/replicate item tree structure into media manager

julien-sebire:
This PR adds the possibility to specify several levels of tree structure to create in the media manager in one call.
The idea behind is to be able to reproduce the tree structure of the item bank in which a set of items with a shared stimulus is imported.
When we import a package with shared stimulus in this place:
item-tree-structure
it creates the similar tree structure to store the stimulus in mediamanager:
media-tree-structure
And finally adds the label of the first item imported as a last tree level.

Requires: 
- https://github.com/oat-sa/generis/pull/763
- https://github.com/oat-sa/extension-tao-mediamanager/pull/151

To test the behaviour, you need:
- a Tao installation with MadiaManager extension
- this sample QTI package with shared stimulus:
[casus_1.zip](https://github.com/oat-sa/extension-tao-itemqti/files/4451617/casus_1.zip)
- go to item bank, create the class(es) that you 'd like to reproduce in MediaManager
- import the package.
It should create 4 items named `Conseiller 001 - Réaction 1` to `Conseiller 001 - Réaction 4`, and in the media manager, the corresponding classes and finally the last class `Conseiller 001 - Réaction 1` named after the label of the first imported item, and in this class, a `shared.xml` file.


model/qti/ImportService.php
```diff
@@ -548,13 +551,19 @@ public function importQtiItem(
                 $itemAssetManager->loadAssetHandler($peHandler);
 
                 if ($this->getServiceLocator()->get(\common_ext_ExtensionsManager::SERVICE_ID)->isInstalled('taoMediaManager')) {
+                    // Media manager path where to store the stimulus.
+                    $path = $this->retrieveFullPathLabels($itemClass);
```
wazelin, line 555: If we reflect a path of an item on import, shouldn't we do the same on a regular media file upload during authoring?

wazelin:
Could you please provide a brief step-by-step guide of importing such an item?
I wasn't able to do replicate the described behavior.

model/qti/ImportService.php
julien-sebire: That's a good question. The specificity of shared stimulus imported from a package where it's used is that the shared stimulus is directly linked to a set of items. So reflecting the path makes sense. When manually uploading an asset, it is not specifically linked to an item, so there's no way to know which tree structure to replicate.

julien-sebire:
About reproducing, I'll send you a sample package with a set of items and shared stimulus so that you can import and see the behaviour.

model/qti/ImportService.php
```diff
@@ -843,4 +852,30 @@ protected function getMetadataImporter()
...
+     * Retrieve the labels of all parent classes up to base item class.
+     *
+     * Return values: an empty string if the $class parameter is not a class object
+     *                an empty array if the $class parameter is the base item class
+     *                an array in other cases, ordered from base item class to the class given as parameter
+     *
+     * @param mixed $class Class from which to retrieve.
+     * @return array|string
+     */
+    public function retrieveFullPathLabels($class)
```
bartlomiejmarszal, line 866: unit test?

julien-sebire: Looking for the possibility to do that before tomorrow.

julien-sebire: Added tests for this method.

PR #1428, opened 2020-04-08
BOSA 81 - Fix size in container for avoid responsive mode affect to answer images

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/BOSA-81

**Description**

The inserted image (in the response alternatives) does not display correctly in the preview.

Precondition: logged in as an admin

**Note: `!important` is necessary to avoid user can overwrite the styles from authoring using responsive mode**

Steps to reproduce:

1. Navigate to items
2. Select the specific item with label name "Test 2 - vb vraag ABS A "
imagen

3. Click on the authoring button
4 Click on the [Save] button
5. Click on the [Preview]

**Actual result**

The inserted image does not display correctly in the response alternatives. See the attached video.

**Expected result**

The inserted image displays correctly in the response alternatives.


wazelin:
Wow, isn't there an `scss` or something?

habemuscode:
Yes sorry, my mistake. I am updating the PR

views/scss/qti/_qti-base.scss
```diff
@@ -217,6 +217,9 @@ $mainContainer : '.qti-item';
 
     .qti-block {
         border:1px solid transparent;
+        & img {
```
wazelin, line 220: You don't need an ampersand here. I mean, it will work, but in fact that's not a correct usage of it.
```suggestion
        img {
```

views/scss/qti/_qti-base.scss
```diff
@@ -217,6 +217,9 @@ $mainContainer : '.qti-item';
...
+            width: inherit !important;
```
wazelin, line 221: And this doesn't fix the issue with the relative size images, for which a `.label-content`, containing a `.qti-block` should have a width of `100%`

habemuscode: With inherit, he will use the value of their parent. I use a circle as responsive with for example 6% and square with a fixed value of ~100px and it works for me
imagen

wazelin: No, it doesn't. The relative size is being set to the `img` itself via `width` attribute. Your so-called relative `6%` is in fact just `width: inherit` and nothing more.
image

wazelin: No, it doesn't. The relative size is being set to the `img` itself via `width` attribute. Your so-called relative `6%` is in fact just `width: inherit` and nothing more.
image

wazelin:
Replaced with #1429.

PR #1430, opened 2020-04-09
Feature/tca 389/item import lock ttl

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TCA-389
 
Added configuration parameter with TTL of the locker for the item import process
 
#### How to test
 
Set configured parameter TTL to the 1, run item importing (item has a lot of media files, that can't be imported in 1 second), check the log that has record, that locker was released before item was imported (so theoretically someone can ask the item files that weren't imported at this time)

#### Dependencies

Companion PR
- [ ] https://github.com/oat-sa/extension-tao-act/pull/1346


model/qti/ImportService.php
```diff
@@ -501,6 +531,7 @@ public function importQtiItem(
                             ) . $validationReport->getMessage()
                         );
                         \common_Logger::i('Item metadata is not valid: ' . $validationReport->getMessage());
+                        $this->checkImportLockTime($startImportTime, $qtiItemResource->getIdentifier());
```
SashaTAO, line 534: I think it would be better to add `final` block and release lock / checkImportLockTime there to avoid duplicated and be safe in case we miss any exception or there will be another return statement

model/qti/ImportService.php
```diff
@@ -84,6 +82,12 @@ class ImportService extends ConfigurableService
      */
     const CONFIG_VALIDATE_RESPONSE_PROCESSING = 'validateResponseProcessing';
 
+    /**
+     * TTL of the item importing process
+     * How long item will be locked while lock service automatically release the lock
+     */
+    const OPTION_IMORT_LOCK_TTL = 'importLockTtl';
```
augustas, line 89: A typo, I guess it should be `OPTION_IMPORT_LOCK_TTL`

model/qti/ImportService.php
zagovorichev: Thank you, @SashaTAO, I thought about that. It's a complex function and requires complex refactoring. I guess we need to create new task for that. If I add final with only these 2 actions it will make this method more messy and unclear.

model/qti/ImportService.php
zagovorichev: Good catch! thank you

SashaTAO:
- [ ] Code is self documented

PR #1437, opened 2020-04-15
BSA-81: Choice interaction image size

wazelin:
- BSA-81 fix(stylesheet): don't stretch a choice-interaction-embedded image to the container's full width by default in order to rely on user-defined values only
- BSA-81 chore(stylesheet): rebuild
- BSA-81 chore(version): bump a fix one

A reproducible case is stored under a `BSA-81` Item on a project-related test environment.


wazelin:
> After checking the item in the ticket, I found that problem not in `width: 100%;`
> Problem in configs, user put size `17%`, but in generated item `17%px`
> image
> image

I see where that confusion may come from, but you're not quite right here. The problem is exactly with `width: 100%` style, which is applied on top of all the others.
The `img` width in `responsive` mode is controlled by `width` attribute on the tag itself, not the `style` (that one should in fact be removed, but it requires a lot more refactoring, which is not really necessary for the fix).

wazelin:
> Please, create a ticket for issue with `17%px`

Way ahead of you :P
It's linked to the previous message.

PR #1447, opened 2020-05-06
csfixer fix

Yury-Razhkou:
Related to: UNO-200
Problem: csfixer removed some needed parts of code.
How to test: import file from task.

### DIFF
--- a/model/qti/templates/feedbacks/rules/qti.choices.tpl.php
+++ b/model/qti/templates/feedbacks/rules/qti.choices.tpl.php
@@ -37,10 +37,11 @@
             <baseValue baseType="identifier"><?=$feedbackIdentifierThen?></baseValue>
         </setOutcomeValue>
     </responseIf>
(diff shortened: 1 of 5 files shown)


model/qti/templates/feedbacks/rules/qti.choices.tpl.php
```diff
@@ -42,5 +42,5 @@
         <setOutcomeValue identifier="<?=$feedbackOutcomeIdentifier?>">
     <baseValue baseType="identifier"><?=$feedbackIdentifierElse?></baseValue>
             </setOutcomeValue>
-<?php endif;?>
+    </responseElse><?php endif;?>
```
wazelin, line 45: Could you please align the opening tag on line 41, so it's more obvious what's going on there?

ktsirkou:
> Related to: UNO-200
> Problem: csfixer removed some needed parts of code.
> How to test: import file from task.

Great catch!!

One remark, we need to remember to restart worker :)

PR #1455, opened 2020-05-14
Provide included elements ids into the ItemUpdatedEvent

gabrielfs7:
Provide to the included elements info for ItemUpdatedEvent

https://oat-sa.atlassian.net/browse/NEX-966

### DIFF
--- a/model/qti/event/UpdatedItemEventDispatcher.php
+++ b/model/qti/event/UpdatedItemEventDispatcher.php
@@ -1,0 +1,68 @@
+<?php
+
+/*
+ * This program is free software; you can redistribute it and/or
+ * modify it under the terms of the GNU General Public License
+ * as published by the Free Software Foundation; under version 2
+ * of the License (non-upgradable).
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ */
+
+namespace oat\taoQtiItem\model\qti\event;
+
+use core_kernel_classes_Resource;
+use oat\oatbox\event\EventManagerAwareTrait;
+use oat\oatbox\service\ConfigurableService;
+use oat\taoItems\model\event\ItemUpdatedEvent;
+use oat\taoQtiItem\model\qti\Item;
+use tao_helpers_Uri;
+
+class UpdatedItemEventDispatcher extends ConfigurableService
+{
+    use EventManagerAwareTrait;
+
+    public function dispatch(Item $qtiItem, core_kernel_classes_Resource $rdfItem): void
+    {
+        $this->getEventManager()->trigger(
+            new ItemUpdatedEvent(
+                $rdfItem->getUri(),
+                [
+                    'includedElementIds' => $this->getIncludedElementIds($qtiItem)
+                ]
+            )
+        );
+    }
+
+    private function getIncludedElementIds(Item $qtiItem): array
+    {
+        $ids = [];
+
+        $elements = (array)($qtiItem->toArray()['body']['elements'] ?? []);
+
+        foreach ($elements as $element) {
+            if ($this->hasIncludedElement($element)) {
+                $ids[] = $this->formatElementId($element['attributes']['href']);
+            }
+        }
+    }
+
+    private function hasIncludedElement(array $element): bool
+    {
+        return ($element['qtiClass'] ?? '') === 'include' && !empty($element['attributes']['href']);
+    }
+
+    private function formatElementId(string $id): string
+    {
+        return tao_helpers_Uri::decode(substr($id, strpos($id, 'http')));
+    }
+}
--- a/model/qti/event/UpdatedItemEventDispatcher.php
+++ b/model/qti/event/UpdatedItemEventDispatcher.php
@@ -18,6 +18,8 @@
  * Copyright (c) 2020 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
  */
 
+declare(strict_types=1);
+
 namespace oat\taoQtiItem\model\qti\event;
 
 use core_kernel_classes_Resource;
--- a/model/qti/event/UpdatedItemEventDispatcher.php
+++ b/model/qti/event/UpdatedItemEventDispatcher.php
@@ -29,13 +29,15 @@ use oat\taoQtiItem\model\qti\parser\IncludedElementIdsExtractor;
 
 class UpdatedItemEventDispatcher extends ConfigurableService
 {
+    private const INCLUDE_ELEMENT_IDS_KEY = 'includeElementIds';
+
     public function dispatch(Item $qtiItem, core_kernel_classes_Resource $rdfItem): void
     {
         $this->getEventManager()->trigger(
             new ItemUpdatedEvent(
                 $rdfItem->getUri(),
                 [
-                    'includedElementIds' => $this->getIncludedElementIdsExtractor()->extract($qtiItem)
+                    self::INCLUDE_ELEMENT_IDS_KEY => $this->getIncludedElementIdsExtractor()->extract($qtiItem)
                 ]
             )
         );
--- a/model/qti/event/UpdatedItemEventDispatcher.php
+++ b/model/qti/event/UpdatedItemEventDispatcher.php
@@ -21,7 +21,7 @@
 namespace oat\taoQtiItem\model\qti\event;
 
 use core_kernel_classes_Resource;
-use oat\oatbox\event\EventManagerAwareTrait;
+use oat\oatbox\event\EventManager;
 use oat\oatbox\service\ConfigurableService;
 use oat\taoItems\model\event\ItemUpdatedEvent;
 use oat\taoQtiItem\model\qti\Item;
@@ -29,8 +29,6 @@ use oat\taoQtiItem\model\qti\parser\IncludedElementIdsExtractor;
 
 class UpdatedItemEventDispatcher extends ConfigurableService
 {
-    use EventManagerAwareTrait;
-
     public function dispatch(Item $qtiItem, core_kernel_classes_Resource $rdfItem): void
     {
         $this->getEventManager()->trigger(
@@ -41,6 +39,11 @@ class UpdatedItemEventDispatcher extends ConfigurableService
                 ]
             )
         );
+    }
+
+    private function getEventManager(): EventManager
+    {
+        return $this->getServiceLocator()->get(EventManager::SERVICE_ID);
     }
 
     private function getIncludedElementIdsExtractor(): IncludedElementIdsExtractor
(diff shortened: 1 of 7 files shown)


model/qti/event/UpdatedItemEventDispatcher.php
```diff
@@ -0,0 +1,50 @@
...
+use core_kernel_classes_Resource;
+use oat\oatbox\event\EventManagerAwareTrait;
+use oat\oatbox\service\ConfigurableService;
+use oat\taoItems\model\event\ItemUpdatedEvent;
+use oat\taoQtiItem\model\qti\Item;
+use oat\taoQtiItem\model\qti\parser\IncludedElementIdsExtractor;
+
+class UpdatedItemEventDispatcher extends ConfigurableService
+{
+    use EventManagerAwareTrait;
```
siwane, line 32: I discourage to use this one

model/qti/event/UpdatedItemEventDispatcher.php
```diff
@@ -0,0 +1,50 @@
...
+
+    public function dispatch(Item $qtiItem, core_kernel_classes_Resource $rdfItem): void
+    {
+        $this->getEventManager()->trigger(
+            new ItemUpdatedEvent(
+                $rdfItem->getUri(),
+                [
+                    'includedElementIds' => $this->getIncludedElementIdsExtractor()->extract($qtiItem)
```
siwane, line 40: I would put this key as public const

model/qti/event/UpdatedItemEventDispatcher.php
siwane, line 32: I discourage to use this one

model/qti/event/UpdatedItemEventDispatcher.php
siwane: also is it possible to put more verbose key, smth like media, or at least include instead of included?

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,53 @@
...
+                $ids[] = $this->formatElementId($element['attributes']['href']);
+            }
+        }
+
+        return $ids;
+    }
+
+    private function hasIncludedElement(array $element): bool
+    {
+        return ($element['qtiClass'] ?? '') === 'include' && !empty($element['attributes']['href']);
```
siwane, line 46: If Im adding an image I have a media uri, should be part of ur scope tasks?
``` </p>```

Comment (gabrielfs7) [reply model/qti/parser/IncludedElementIdsExtractor.php:76]: This was split to another task: https://oat-sa.atlassian.net/browse/NEX-1066

File: model/qti/event/UpdatedItemEventDispatcher.php (hunk 1 above)
Comment (gabrielfs7) [reply model/qti/event/UpdatedItemEventDispatcher.php]: Ok, used specific one.

File: model/qti/event/UpdatedItemEventDispatcher.php (hunk 2 above)
Comment (gabrielfs7) [reply model/qti/event/UpdatedItemEventDispatcher.php]: I kept as a private const (only used here, so no need for external usage). I changed to `includeElementIds` cause the ids belong to a include in the item, which could be many things and the client can decide what to do.

File: model/qti/parser/IncludedElementIdsExtractor.php [hunk 4]
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    private function formatElementId(string $id): string
+    {
+        return tao_helpers_Uri::decode(substr($id, strpos($id, 'http')));
```
boajer, line 86: Are we limited here to `http` only (whatever  it stands for ) ?

gabrielfs7: As far as I could see we always use this URI regardless storage. So the item content will have something like this: 

```
...
"xinclude_5ec29c5c76661791536613": {
                    "serial": "xinclude_5ec29c5c76661791536613",
                    "qtiClass": "include",
                    "attributes": {
                        "href": "taomedia:\/\/mediamanager\/https_2_test-tao-deploy_0_docker_0_localhost_1_ontologies_1_tao_0_rdf_3_i5ec293a38ebe623833180e3b0a547a6d4"
                    },
...
```

So for the shared stimulus purpose it works. DO you know how it is possible to have something different on that situation? 

The other media relations will be covered on other task.

model/qti/event/UpdatedItemEventDispatcher.php
siwane: <xi:nclude> can only be a sharedstimulus, e.q. any xml. If we have only one media, it is not using xinlude at all. Its why Im asking if we have to take care about these medias

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,88 @@
...
+        $this->ids = [];
+
+        $this->incrementIds($this->getItemBody($qtiItem));
+
+        return array_unique($this->ids);
+    }
+
+    private function getItemBody(Item $qtiItem): array
+    {
+        return json_decode(json_encode($qtiItem->toArray()['body']), true);
```
mike-ionut-mihai-sandu-tao, line 45: this looks a little bit weird

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,88 @@
...
+    /** @var string[] */
+    private $ids;
+
+    public function extract(Item $qtiItem): array
+    {
+        $this->ids = [];
+
+        $this->incrementIds($this->getItemBody($qtiItem));
+
+        return array_unique($this->ids);
```
mike-ionut-mihai-sandu-tao, line 40: does this mean we will drop some versions of the value with different keys ?

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,88 @@
...
+    private function incrementIds(array $body): void
+    {
+        foreach ($body as $key => $value) {
+            $this->incrementIdsByKey($key, $value);
+        }
+    }
+
+    private function incrementIdsByKey(string $key, $value): void
+    {
+        if (!is_array($value)) {
```
mike-ionut-mihai-sandu-tao, line 57: also i see we are dropping non array value, so maybe this is processing and not a incrementIds

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: The ids are unique for media (sharedStimulus) this is the URI id used as reference. The array_unique is to avoid repeat the same id twice.

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: Yes, all the methods are processing, but the outcome is to have the ids incremented according to some criteria, in this case based on the key.

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: This is the way to avoid objects, so forcing the object to be an array and work as an array only, otherwise I would have many stdClass + array mixed which makes the code more complex. So I normalise the data to array before start.

model/qti/event/UpdatedItemEventDispatcher.php
gabrielfs7: I see. Now we only care about shared stimulus, but other media will be handled in other task, so we will probably have another parameter `mediaIds`. I would prefer keep then separately in the event payload cause we can have different actions with them.

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: I am using TaoMediaResolver as we discussed.

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: Method renamed. Thanks!

mike-ionut-mihai-sandu-tao:

mike-ionut-mihai-sandu-tao:
before merging this let's have a deployment with the branches so we can confirm with QA that this does not affect our customers default behavior

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,102 @@
...
+
+            return;
+        }
+
+        foreach ($value as $element) {
+            $this->incrementIdsByElement((array)$element);
+        }
+    }
+
+    private function incrementIdsByElement(array $element): void
```
siwane, line 75: Are u going accross all item elements or only accross media elements?

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,102 @@
...
+        $id = $this->getMediaResolver()
+            ->resolve($url)
+            ->getMediaIdentifier();
+
+        return tao_helpers_Uri::decode($id);
+    }
+
+    private function getMediaResolver(): TaoMediaResolver
+    {
+        return $this->getOption(self::OPTION_MEDIA_RESOLVER) ?? new TaoMediaResolver();
```
siwane, line 100: Where are u defining that this service is this configuration? u miss to configure at update and install

gabrielfs7: I wanted to make it configurable for tests only, cause `TaoMediaResolver` does not extend ConfigurableService. 

What is the best option? 

- TaoMediaResolver extends ConfigurableService
- TaoMediaResolver at the Update script?

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: Across all the elements that contain `elements` key. It was needed, cause some cases we have sharedStimulus nested to other elements. What do you suggest?

model/qti/parser/IncludedElementIdsExtractor.php
siwane: > TaoMediaResolver extends ConfigurableService
Cannot because children are using construct args (vs configurableService $options)
> TaoMediaResolver at the Update script?
Why not, and at install. It will registered EventDispatcher with associated resolver config
> new TaoResolver()
It is not the best solution, but it requires minimal effort and do not impact actual behavior

Welcome in TAO dependencies injections

model/qti/parser/IncludedElementIdsExtractor.php
siwane: need to investigate, to use Item parser

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: Ok, I decided to use the third option, cause I do not want to make it configurable by user as option if it is not needed. I used DI pattern. Thanks!!

model/qti/parser/IncludedElementIdsExtractor.php
gabrielfs7: Followed your approach! Thanks @siwane . Updated!

model/qti/parser/IncludedElementIdsExtractor.php
```diff
@@ -0,0 +1,69 @@
...
+                ->resolve($element->attr('href'))
+                ->getMediaIdentifier();
+
+            $this->ids[] = tao_helpers_Uri::decode($id);
+        }
+
+        return array_unique($this->ids);
+    }
+
+    public function withMediaResolver(TaoMediaResolver $mediaResolver): self
```
siwane, line 54: nice

model/qti/event/UpdatedItemEventDispatcher.php
```diff
@@ -0,0 +1,55 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2020 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+ */
+
```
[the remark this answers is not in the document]
siwane: up

PR #1482, opened 2020-07-02
Fix/nex 1227/fix broken unit test

bartlomiejmarszal:
Cause of following error triggered in my unit test in [PR](https://github.com/oat-sa/extension-tao-mediamanager/pull/212) I changed broken class that was using const that was created after tao instance installation. 

PSR-12 applied

### DIFF
--- a/model/qti/Element.php
+++ b/model/qti/Element.php
@@ -754,9 +754,4 @@ abstract class Element implements Exportable
     {
         return $this->getServiceLocator()->get(ApplicationService::SERVICE_ID);
     }
-
-    private function getExtensionManager(): common_ext_ExtensionsManager
-    {
-        return $this->getServiceLocator()->get(common_ext_ExtensionsManager::SERVICE_ID);
-    }
 }
--- a/model/qti/Element.php
+++ b/model/qti/Element.php
@@ -1,20 +1,43 @@
 <?php
 
-/*
- * This program is free software; you can redistribute it and/or modify it under the terms of the GNU General Public License as published by the Free Software Foundation; under version 2 of the License (non-upgradable). This program is distributed in the hope that it will be useful, but WITHOUT ANY WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the GNU General Public License for more details. You should have received a copy of the GNU General Public License along with this program; if not, write to the Free Software Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA. Copyright (c) 2013 (original work) Open Assessment Technologies SA (under the project TAO-PRODUCT);
+/**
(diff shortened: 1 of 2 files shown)


mike-ionut-mihai-sandu-tao:

model/qti/Element.php
```diff
@@ -1,20 +1,43 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
+ *
+ * Copyright (c) 2013-2020 (original work) Open Assessment Technologies SA;
  */
 
+declare(strict_types=1);
```
gabrielfs7, line 21: Good job improving the code :)

model/qti/Element.php
```diff
@@ -126,10 +161,10 @@ public function removeAttributeValue($name)
      * Set the attributes for the the Qti Element
      * Argument format: array(attributeName => value)
      *
-     * @param array $values
-     * @throws InvalidArgumentException
+     * @throws AttributeException
+     * @throws QtiModelException
      */
-    public function setAttributes($values)
+    public function setAttributes(array $values): void
```
gabrielfs7, line 167: This might break, unfortunately :/ 

Cause internally this method checks is parameter is an array, so typehint is dangerous. Can you please check the risk to add this?

model/qti/Element.php
```diff
@@ -282,23 +319,23 @@ public function getIdentifiedElement($identifier, $elementClasses = [])
...
-     * @param string $name
      * @param mixed $value
+     *
      * @return mixed
+     *
+     * @throws QtiModelException
+     * @throws AttributeException
      */
-    public function attr($name, $value = null)
+    public function attr(string $name, $value = null)
```
gabrielfs7, line 338: This is nullable. It is better to double check the methods internals to see if the typehints will not break.

model/qti/Element.php
```diff
@@ -595,10 +620,8 @@ public function getComposingElements($className = '')
 
     /**
      * Get the Qti Item the current Qti Element belongs to
-     *
-     * @return \oat\taoQtiItem\model\qti\Item
      */
-    public function getRelatedItem()
+    public function getRelatedItem(): Item
```
gabrielfs7, line 624:
```suggestion
    public function getRelatedItem(): ?Item
```

This is nullable

model/qti/Element.php
bartlomiejmarszal: I was thinking that I should even remove `else` from this method and keep parameter typed.

model/qti/Element.php
bartlomiejmarszal: It's not used without param anywhere in `usages` list

PR #1491, opened 2020-07-15
Feature/nex 1135/media item relation migration

boajer:
https://oat-sa.atlassian.net/browse/NEX-1135


model/qti/event/UpdatedItemEventDispatcher.php
```diff
@@ -61,4 +55,15 @@ private function getElementReferencesExtractor(): ElementReferencesExtractor
...
+    public function extractReferences(Item $qtiItem): array
+    {
+        $extractor = $this->getElementReferencesExtractor();
+
+        return [
+            self::INCLUDE_ELEMENT_REFERENCES_KEY => $extractor->extract($qtiItem, XInclude::class, 'href'),
+            self::OBJECT_ELEMENT_REFERENCES_KEY => $extractor->extract($qtiItem, QtiObject::class, 'data'),
+            self::IMG_ELEMENT_REFERENCES_KEY => $extractor->extract($qtiItem, Img::class, 'src'),
+        ];
+    }
```
gabrielfs7, line 68: This class is an `Event dispatcher`. "Extract" is not part of his responsibility. Can you please move it to the `ElementReferencesExtractor` class instead with a proper method name? I would suggest:

1. Create a new VO to represent the element references:

```
class ElementReferences
{
    /** @var string[] */
    private $xIncludeReferences;

    /** @var string[] */
    private $objectReferences;

    /** @var string[] */
    private $imgReferences;
    
    public function __construct(array $xIncludeReferences, array $objectReferences, array $imgReferences)
    {
    }

    public function getXIncludeReferences(): array
    {
        return $this->xIncludeReferences;
    }

    public function getObjectReferences(): array
    {
        return $this->objectReferences;
    }

    public function getImgReferences(): array
    {
        return $this->imgReferences;
    }
}
```

2. Create the new method in the extractor:

```
class ElementReferencesExtractor extends ConfigurableService
{
    // ... keep previous method

    /**
     * @throws Exception
     */
    public function extractAll(Item $qtiItem): ElementReferences
    {
        return new ElementReferences(
            $this->extract($qtiItem, XInclude::class, 'href'),
            $this->extract($qtiItem, QtiObject::class, 'data'),
            $this->extract($qtiItem, Img::class, 'src')
        );
    }
}
```

3. Use the new method in the `UpdatedItemEventDispatcher.php` keeping the constants there.

This way you are not exposing specific event payload keys externally.

model/qti/event/UpdatedItemEventDispatcher.php
```diff
@@ -61,4 +55,15 @@ private function getElementReferencesExtractor(): ElementReferencesExtractor
     {
         return $this->getServiceLocator()->get(ElementReferencesExtractor::class);
     }
+
```
siwane, line 58: Can we extract this logic to a dedicated service?

PR #1494, opened 2020-07-20
Fix - NEX 1189 - [FE] Shared Stimulus: Picture gets smaller on canvas when user quits Image properties

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/NEX-1189

**Description**

When the user adds image, changes its size to any but 100% while Responsive mode is ON, the picture gets smaller as soon as the user quits Image properties. 
However, the widths remains the same (height is changed) but as far as the picture got smaller, the gray area appears on the right side of the picture to full fill the remained  width.
NOTE: issues doesn't reproduce if size is set to 100%, or if Responsive mod is OFF.

**Steps to reproduce**

- Open a Shared Stimulus in Authoring mode
- Add image by clicking 'Insert image' button
- Verify the 'Response mode' checkbox is checked
- Change the image size to any but 100%
- Quit Image Properties by clicking any area inside stimulus block

**Actual result** 

Image got smaller and the gray area appeared on the right side of it.

**Expected result**

Image remains of the same size as was set in Image Properties.

imagen


views/scss/item-creator.scss
```diff
@@ -789,6 +789,7 @@ img.dragging.ui-draggable-dragging {
         .widget-box {
             &.edit-sleep {
                 position: relative;
+                width: 100% !important;
```
wazelin, line 792: That will break a `Size and positioning` feature in authoring mode, won't it?
image

habemuscode: You tested in your local?

wazelin: I'm asking you to kindly check that if it hasn't been done already.

habemuscode: As I said in DSM, from the list of all possible solutions I discover, this is the best. Don't break animation but still fix the bug when you click outside and works for `items` and `mediaManager`. If you see some unexpected behavior tell me

wazelin: Well, the `span` itself then stretches out, as it shouldn't, and the picture changes size back and forth when you click on it.

Try the following instead.
```suggestion

                img {
                    min-width: 100%;
                }

```

habemuscode:
@wazelin Your solution avoids that 1s change size, good job :)

PR #1495, opened 2020-07-21
[NEX-1456] Add possibility to select custom response processing in test item authoriding

atsymuk:
Related to : https://oat-sa.atlassian.net/browse/NEX-1456
 
Add possibility to select custom response processing type in case if [xmlEditRp](https://github.com/oat-sa/extension-tao-xmledit-responseprocessing) installed
  
#### How to test
 
- prepare an instance of TAO with xmlEditRp extension
- create an item
- navigate to item authoring
- try to set up response processing for the interaction
- make sure that it is possible to select custom response processing
- after selected custom response processing make sure that all the interaction switch to this type
- try to add an interaction and make sure that newly added interactions has custom response processing
- check the default value of custom response processing


atsymuk:
> Regarding the code, it looks ok. Just one comment about a deprecated API.
> 
> Regarding the feature, yes it adds the `custom` processing mode. However, we lose the existing processing scripts, which is not good IMO. We should at least have a warning dialog before losing any script. But a good approach would have been to keep the existing script, eventually replacing the template line by its actual script.

now the default value for custom RP pre populated based on previous PR

views/js/qtiCreator/model/variables/ResponseDeclaration.js
```diff
@@ -13,8 +13,8 @@ define([
     _.extend(methods, editable);
     _.extend(methods, {
         setTemplate : function(template){
-            var templateUri = responseHelper.getTemplateUriFromName(template);
-            if(templateUri && this.template !== templateUri){
+            var templateUri = responseHelper.getTemplateUriFromName(template) || null;
+            if(this.template !== templateUri){
```
bziondik, line 17: Why was removed condition `templateUri && `?

atsymuk: because for custom response type there is no `templateUri` so it will be changed to `null`

PR #1498, opened 2020-07-27
Code styles fixes in feedback implementation

Alroniks:
I found out it when I worked on [BSA-104](https://oat-sa.atlassian.net/browse/BSA-104).

"Use" statements in these files includes classes that already defined in that namespace, it may trigger unexpected errors in the future. Also, I fixed some other issues, mostly related to code style and common sense.

### DIFF
--- a/model/qti/feedback/Feedback.php
+++ b/model/qti/feedback/Feedback.php
@@ -96,10 +96,7 @@ abstract class Feedback extends IdentifiedElement implements FlowContainer, Cont
         $collection = $relatedItem->getIdentifiedElements();
 
         try {
-            $feedback = $collection->getUnique($newIdentifier, self::class);
-            if (is_null($feedback)) {
-                return true;
-            }
+            return null === $collection->getUnique($newIdentifier, self::class);
         } catch (QtiModelException $e) {
         }
 
--- a/model/qti/feedback/Feedback.php
+++ b/model/qti/feedback/Feedback.php
@@ -1,6 +1,5 @@
 <?php
-
-/*
+/**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * as published by the Free Software Foundation; under version 2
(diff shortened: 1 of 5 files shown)


model/qti/feedback/ModalFeedback.php
```diff
@@ -47,18 +41,14 @@ class ModalFeedback extends Feedback
...
         );
     }
     
     public function toForm()
     {
-        $formContainer = new ModalFeedback($this);
-        return $formContainer->getForm();
+        return (new self($this))->getForm();
     }
 }
```
shpran, line 54: Are we using this method?
And I think that this one won't work, its parents `__construct` expects an array, we will get the error here: https://github.com/oat-sa/extension-tao-itemqti/blob/ee1d800d983e1cb2bac8fc0e7dbb5b2465c36877/model/qti/Element.php#L139

model/qti/feedback/Feedback.php
```diff
@@ -90,7 +95,7 @@ public function isIdentifierAvailable($newIdentifier)
                 $collection = $relatedItem->getIdentifiedElements();
 
                 try {
-                    $feedback = $collection->getUnique($newIdentifier, 'oat\\taoQtiItem\\model\\qti\\feedback\\Feedback');
+                    $feedback = $collection->getUnique($newIdentifier, static::class);
```
wazelin, line 98: I don't really know if it makes a difference, but just to be safe, kindly consider replacing the `static` reference with `self`, in order to keep the behavior as it was.
```suggestion
                    $feedback = $collection->getUnique($newIdentifier, self::class);
```

model/qti/feedback/ModalFeedback.php
Alroniks: I am not sure that it is actually used somewhere, but probably yes. I am fixed only code styles and issues, that IDE highlighted by red across all files. But yes, it makes sense to remove it as unused.

model/qti/feedback/Feedback.php
Alroniks: Actually it does not matter. But I don't mind.

model/qti/feedback/ModalFeedback.php
Alroniks: I removed the method as unused.

model/qti/feedback/Feedback.php
```diff
@@ -59,28 +65,27 @@ public function getBody()
...
         $returnValue = false;
 
         if (empty($newIdentifier) || is_null($newIdentifier)) {
             throw new InvalidArgumentException("newIdentifier must be set");
         }
 
-        if (!empty($this->identifier) && $newIdentifier == $this->identifier) {
+        if (!empty($this->identifier) && $newIdentifier === $this->identifier) {
             $returnValue = true;
         } else {
```
wazelin, line 90: Let's get rid of this ugly `else` block by introducing an early return, based on the condition above.

Alroniks:
@wazelin I simplified code flow

PR #1499, opened 2020-07-29
Fix/BSA-202/Check the base64 encoded image

shpran:
Relates to: [BSA-202](https://oat-sa.atlassian.net/browse/BSA-202)

**Changes:**
* Added check for the base64 encoded image.

**Require:**
* [#2621](https://github.com/oat-sa/tao-core/pull/2621)

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -43,11 +43,11 @@ return [
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.4.3',
+    'version'     => '25.5.0',
     'author'      => 'Open Assessment Technologies',
     'requires' => [
         'taoItems' => '>=10.8.2',
-        'tao'      => '>=45.1.3',
+        'tao'      => '>=45.2.0',
         'generis'  => '>=12.17.0',
     ],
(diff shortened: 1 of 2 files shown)


manifest.php
```diff
@@ -43,11 +43,11 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.4.2',
+    'version'     => '25.4.3',
```
wazelin, line 46: I believe we should bmp at least the minor version here (in fact, I would even consider bumping a major one), the PR introduces a major dependency update, which on its own may be a breaking change.

manifest.php
```diff
@@ -43,11 +43,11 @@
...
     'author'      => 'Open Assessment Technologies',
     'requires' => [
         'taoItems' => '>=10.8.2',
-        'tao'      => '>=43.1.0',
+        'tao'      => '>=45.1.3',
```
wazelin, line 50: Just a friendly reminder, in case you decide to accept [that suggestion](https://github.com/oat-sa/tao-core/pull/2621#discussion_r462253756).

Don't forget to update the dependency version here as well.

PR #1506, opened 2020-08-06
Fix/double xml declaration tag pci markup

bugalot:
### Description

We discovered an issue in Sprint 133 and above. There is no more way to build a Delivery based on a Test containing items having PCI textReaderInteractions (and maybe other PCIs). This affects all customers from Sprint 133 + Large-Scale campaings operations. **In other words this is quite important.**

Screenshot_2020-08-07_00-50-31

**We are not confident with this fix**, because we have no idea what impact it can have on other parts of the system like Shared Stimulus, Passages, MediaManager, PCI Authoring and co... **We kindly ask all the people knowledgeable in these areas to jump on this** to give their advice about whether or not this could impact other parts of the system. 

**Please review me!**

### Steps to Reproduce

1. Create an Item with a PCI textReaderInteraction. Save it.
2. Create a Test with the Item created at step 1.
3. Compile the Test created at step 2 as a Delivery.

The compilation fails.

### Additional Information

* After long debugging sessions, we discovered that the issue is due to the final item markup generated at Delivery compilation time. Multiple `<?xml version="1.0"?>` XML Declaration Tags can be found across the item (in `<markup/>` elements), making the XML unable to be loaded via `DOMDocument`. That's why we suppress them in this Pull Request.

* It is interesting to see that this does not happen at export time. Only at Delivery Compilation time.


model/qti/XIncludeLoader.php
```diff
@@ -133,7 +133,7 @@ private function parseCustomElementMarkup(CustomInteraction $customElement)
             throw new ParsingException('cannot parse pci markup');
         }
 
-        $customElement->setMarkup($xml->saveXML(null, LIBXML_NOEMPTYTAG));
+        $customElement->setMarkup($xml->saveXML($xml->documentElement, LIBXML_NOEMPTYTAG));
```
tikhanovichA, line 136: Also, why we need to save it even if the `if ($loadSuccess && !is_null($node)) {` is false?
Maybe we can add a check if `$xincludesNodes->count()` then do some things and save it.
What do you think?

jbout:
Even before the bug appeared this function returned the entire XML, so I'm skeptical, because this fix could have negative effects across the system. Having a unit test that highlights the issue would be very helpfull

model/qti/XIncludeLoader.php
mike-ionut-mihai-sandu-tao: just tested and if the package does not contain shared-stimulus or pci , it will not trigger that path of the code,
and while verifying the value of the `$xincludesNodes->count()` it is allays =

mike-ionut-mihai-sandu-tao:

gabrielfs7:

PR #1513, opened 2020-08-13
Fix TAO 10288 - The 'Properties' section becomes partly hidden after setting modal feedback options

habemuscode:
**Related to this issue**

https://oat-sa.atlassian.net/browse/TAO-10288

**Description**

**ONLY ON CHROME, MS EDGE and IE11**

Whenever the user is trying to change the modal feedback 'if' setting from correct to any other option, the whole Properties section shifts to the left and becomes partly overlapped by the canvas.

**Preconditions**

An item created with any interaction in it (in which modal feedback can be set).

**Steps to reproduce**

- Open created item in authoring mode.
- Select contained interaction and go to the Response section.
- In the Response properties on the right side of the screen for Modal feedback change 'if' option from 'correct' to any other.

**Actual result**

The whole Properties section becomes partly hidden by the canvas.

**Expected result**

There are no visual changes happens with Properties section (except for changing 'if' option).


views/scss/qtiCreator/_outcome-editor.scss
```diff
@@ -79,4 +79,9 @@
             }
         }
     }
+}
+
+// For solve issue in Chrome with scroll: https://stackoverflow.com/questions/22898335/strange-behavior-of-overflow-auto-on-chrome
+.item-editor-sidebar #sidebar-right-response-properties {
```
wazelin, line 85: Does it really needed to nest an ID selector under another one?
```suggestion
#sidebar-right-response-properties {
```

habemuscode:
Sorry @wazelin , forgot.

PR #1547, opened 2020-09-24
Feature/nex 1817/svgz support

boajer:
https://oat-sa.atlassian.net/browse/NEX-1817

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -43,7 +43,7 @@ return [
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.12.1',
+    'version'     => '25.13.0',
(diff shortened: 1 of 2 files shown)


manifest.php
```diff
@@ -43,7 +43,7 @@
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.12.1',
+    'version'     => '25.13.0',
     'author'      => 'Open Assessment Technologies',
     'requires' => [
```
mike-ionut-mihai-sandu-tao, line 48: do we need to bump the version for the `returnStream` change?

mike-ionut-mihai-sandu-tao (commented):
good job, i think we need to update the dependency version

PR #1554, opened 2020-10-01
[Fix] TAO 10544 - Several responses text entry block

habemuscode:
**Related to**

https://oat-sa.atlassian.net/browse/TAO-10544

**Description**

After setting the map response function to the Inline Text Entry, the added options can disappear in case if the user copies the text from any of the existing ones and adds it to the new option (in this case the elder option will disappear). 

Preconditions: a created item with Inline Text Entry interaction and set map response mode in the properties;

**Steps to reproduce**

1. Open the item and move to the Response tab;
2. Click 'Add a new option' and enter any value there (string); repeat for several times;
3. Add one more option, copy the value from any of the existing options, paste it to the newly added, and edit the entered value (e.g. add a number to the text)
3. Click the Done button;
4. Move to the Response tab again, focus on the option the value of which was copied from.

**Actual result**

The system recognizes the existing option as a clone and removes it.

**Expected result**

As an author I just add a new response item, I expect it to be added, and not replace one of the pre-existing ones, even if the first characters of the name are identical.


views/js/qtiCreator/widgets/interactions/textEntryInteraction/states/Map.js
```diff
@@ -56,7 +56,7 @@ define([
                     score = parseInt($tr.find('.score').val());
 
                 //@todo trigger validate score to check if is a valid score
-                if(!isNaN(score)){
+                if(!isNaN(score) && !(value in response.mapEntries)){
```
andrey-ponamarev, line 59: The `in` operator is going to check property in the prototype chain as well 
```suggestion
                if(!isNaN(score) && !(response.mapEntries.hasOwnProperty(value)){
```

andrey-ponamarev (changes requested):

atsymuk (approved):

bziondik (approved):
- [ ]  New code is covered by tests (if applicable)
- [x]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  New code is respecting code style rules
- [x]  New code is respecting best practices
- [x]  New code is not subject to concurrency issues (if applicable)
- [x]  Feature is working correctly on my local machine (if applicable)
- [x]  Acceptance criteria are respected
- [x]  Pull request title and description are meaningful

PR #1565, opened 2020-10-19
Fix/tao 10562/select point arrowskeys in response

zagovorichev:
Pull request summary
 
Related to : https://oat-sa.atlassian.net/browse/TAO-10562
 
**Select Point** interaction. When focus in the Response -> Score Input - it should use arrows as a input controls (without the fix it uses arrows as a shape control - to move its position)
 
#### Steps to reproduce  (for bugs)
1. Login as admin/item author
2. Go to Items tab and create a new item
3. Go to Authoring and add Select point graphic interaction to the canvas
4. Go to Response and add a spot
5. Click on score value 
6. Push and hold any KB arrow
  
#### How to test
 
Try to do the same and note that arrows now controls cursor position in the input

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -43,7 +43,7 @@ return [
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.15.2',
+    'version'     => '25.15.3',
     'author'      => 'Open Assessment Technologies',
     'requires' => [
         'taoItems' => '>=10.13.0',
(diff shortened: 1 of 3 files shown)


atsymuk (changes requested):
it is possible to use arrow keys inside score input. but if input is not focused, arrow keys still move spot and also duplicate score value while moving. And also one more problem if after activation the spot was moved and activated again there will be multiple score popups

zagovorichev:
@atsymuk Thank you, I've missed this part. Check it please, one more time, I fixed this issue.

atsymuk (approved):

PR #1571, opened 2020-10-28
fix/UNO-116-checkbox-overlaping

meefox:
This is 4 of 4 parts that is needed for https://oat-sa.atlassian.net/browse/TR-43 (ex UNO-116)
here is an important part that contains scss changes, as well as some selectors fixes
P.S. package.json will be committed next

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -43,7 +43,7 @@ return [
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.15.2',
+    'version'     => '25.15.3',
     'author'      => 'Open Assessment Technologies',
     'requires' => [
         'taoItems' => '>=10.13.0',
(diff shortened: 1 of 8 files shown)


btamas:
I still have the issue
Screen Shot 2020-10-29 at 12 19 12

jsconan (changes requested):
Thats a lot of changes in the layout, which surely represent a lot of work.
But this may break some other features.
I did not check all, but it may have impact on keyboard navigation and some other features in the test runner. Is-it not possble to fix the issue without such a change? I'm afraid it will break in some places we didn't think of.

The more visible issues I saw was actually the layout looks a bit broken in the authoring:
before: 
image

after: 
image

When creating the item, in question mode, we can tick the radio/checkboxes. But we should not.
image

Additional remarks:
- as this PR has a dependency to the extension `tao`, it should also reflect this in the `manifest.php` file, in the `requires` section.
- the SCSS should be compiled and the changes added to the PR as well
    ```
    cd tao/views/build
    npm grunt taoqtiitemsass
    ```

meefox:
Please check an alternative solution https://github.com/oat-sa/extension-tao-itemqti/pull/1572

PR #1596, opened 2020-12-04
add all required media player dependencies to portable lib

meefox:
- [ ]  version bump


meefox:
- [ ] version bump

meefox:
outdated

PR #1586, opened 2020-11-19
Fix/ms 579/highlighter support in text entry inline interaction

AnastasiaArcadia:
Related to: https://oat-sa.atlassian.net/browse/MS-579

**Changes**

In text entry inline interactions `input` element was changed to `div` for `reviewRenderer` to support highlight plugin.

**How to test?**

- Create a test with text entry inline interaction
- Take the test on tao
- Login as a scorer on MS to review the responses
- All the text entry inline interactions should be displayed as expected and should have been rendered in a `div` instead of `input`

**Depended on**
- [x] Release and merged https://github.com/oat-sa/tao-item-runner-qti-fe/pull/121

PR #1598, opened 2020-12-11
Fix/ms 532/update item runner fe

anshul119:
Ticket: https://oat-sa.atlassian.net/browse/MS-532

- Updated item runner FE


andrey-ponamarev (changes requested):
Please update version in manifest

andrey-ponamarev:
Version is updated in https://github.com/oat-sa/extension-tao-itemqti/pull/1586

PR #1603, opened 2020-12-21
Fix/CON-144/Booklet preview fails due to timeout

shpran:
Relates to: [CON-144](https://oat-sa.atlassian.net/browse/CON-144)

**Changes:** changed the order of item attributes.

**Companion PR:** [#2760](https://github.com/oat-sa/tao-core/pull/2760)

### DIFF
--- a/model/qti/Item.php
+++ b/model/qti/Item.php
@@ -175,12 +175,12 @@ class Item extends IdentifiedElement implements FlowContainer, IdentifiedElement
     {
         return [
             'oat\\taoQtiItem\\model\\qti\\attribute\\Title',
-            'oat\\taoQtiItem\\model\\qti\\attribute\\Label',
-            'oat\\taoQtiItem\\model\\qti\\attribute\\Lang',
             'oat\\taoQtiItem\\model\\qti\\attribute\\Adaptive',
             'oat\\taoQtiItem\\model\\qti\\attribute\\TimeDependent',
+            'oat\\taoQtiItem\\model\\qti\\attribute\\Lang',
+            'oat\\taoQtiItem\\model\\qti\\attribute\\Label',
             'oat\\taoQtiItem\\model\\qti\\attribute\\ToolName',
-            'oat\\taoQtiItem\\model\\qti\\attribute\\ToolVersion'
+            'oat\\taoQtiItem\\model\\qti\\attribute\\ToolVersion',
         ];
     }
 
--- a/model/qti/ParserFactory.php
+++ b/model/qti/ParserFactory.php
@@ -1528,16 +1528,14 @@ class ParserFactory
         return $returnValue;
     }
 
-    private function buildStylesheet(DOMElement $data)
+    private function buildStylesheet(DOMElement $data): Stylesheet
     {
-        $returnValue = new Stylesheet([
-            'href' => (string) $data->getAttribute('href'),
-            'title' => $data->hasAttribute('title') ? (string) $data->getAttribute('title') : '',
-            'media' => $data->hasAttribute('media') ? (string) $data->getAttribute('media') : 'screen',
-            'type' => $data->hasAttribute('type') ? (string) $data->getAttribute('type') : 'text/css',
+        return new Stylesheet([
+            'href' => $data->getAttribute('href'),
+            'type' => $data->hasAttribute('type') ? $data->getAttribute('type') : 'text/css',
+            'media' => $data->hasAttribute('media') ? $data->getAttribute('media') : 'screen',
+            'title' => $data->getAttribute('title'),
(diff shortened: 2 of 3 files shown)


model/qti/Item.php
```diff
@@ -175,12 +175,12 @@ protected function getUsedAttributes()
     {
         return [
             'oat\\taoQtiItem\\model\\qti\\attribute\\Title',
-            'oat\\taoQtiItem\\model\\qti\\attribute\\Label',
-            'oat\\taoQtiItem\\model\\qti\\attribute\\Lang',
             'oat\\taoQtiItem\\model\\qti\\attribute\\Adaptive',
             'oat\\taoQtiItem\\model\\qti\\attribute\\TimeDependent',
+            'oat\\taoQtiItem\\model\\qti\\attribute\\Lang',
```
gabrielfs7, line 180: @shpran Can you explain why this change was needed, please?

shpran: When we are trying to update (save) item - we validate DOM content before processing and `<assesmentItem>` node attributes order is:
* adaptive
* timeDependent
* xml:lang
* label
**Example:**
```
<assessmentItem
  xmlns="http://www.imsglobal.org/xsd/imsqti_v2p2"
  xmlns:m="http://www.w3.org/1998/Math/MathML"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.imsglobal.org/xsd/imsqti_v2p2 http://www.imsglobal.org/xsd/qti/qtiv2p2/imsqti_v2p2.xsd"
  identifier="i5ea29a5d4e119112935c8357c06037d389"
  title="Aperçu"
  adaptive="false"
  timeDependent="false"
  xml:lang="fr-FR"
  label="2 Aperçu"
  toolName="TAO"
  toolVersion="3.4.0-sprint143"
>
```

When we are trying to preview item/booklet/etc. - we validate processed DOM content, and in this case the order is:
* label
* xml:lang
* adaptive
* timeDependent
**Example:**
```
<assessmentItem
  xmlns="http://www.imsglobal.org/xsd/imsqti_v2p2"
  xmlns:m="http://www.w3.org/1998/Math/MathML"
  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
  xsi:schemaLocation="http://www.imsglobal.org/xsd/imsqti_v2p2 http://www.imsglobal.org/xsd/qti/qtiv2p2/imsqti_v2p2.xsd"
  identifier="i5ea29a5d4e119112935c8357c06037d389"
  title="Aperçu"
  label="2 Aperçu"
  xml:lang="fr-FR"
  adaptive="false"
  timeDependent="false"
  toolName="TAO" toolVersion="3.4.0-sprint143"
>
```

To avoid unnecessary validation - I'm trying to compare stored content with provided. But in our case they will be fully similar except this attributes order. 
In this case we need to fix this order to make stored content and validated content similar.

gabrielfs7: I got it now! Thanks!

gabrielfs7: @shpran Another question about this. Seems that the order of the elements is really important and can cause issues as you described. 

Do not you think this should be explicitly declared so any new attribute added or refactoring on this code will not break it again? (I can say for instance some developer might be tempted to order these fields in alphabetical order - which will break again).

I mean, like a specialised common validation that sort the fields before validating, so it is explicit that the order matters.

I see you fix the problem here, but it wont avoid the same problem in the future. Wdyt?

gabrielfs7 (commented):
I think we should make the elements order importance declarative. I left another comment about that.

model/qti/ParserFactory.php
```diff
@@ -1528,16 +1528,14 @@ protected function getNonEmptyChildren(DOMElement $data)
...
+    private function buildStylesheet(DOMElement $data): Stylesheet
     {
-        $returnValue = new Stylesheet([
-            'href' => (string) $data->getAttribute('href'),
-            'title' => $data->hasAttribute('title') ? (string) $data->getAttribute('title') : '',
-            'media' => $data->hasAttribute('media') ? (string) $data->getAttribute('media') : 'screen',
-            'type' => $data->hasAttribute('type') ? (string) $data->getAttribute('type') : 'text/css',
+        return new Stylesheet([
+            'href' => $data->getAttribute('href'),
+            'type' => $data->hasAttribute('type') ? $data->getAttribute('type') : 'text/css',
```
gabrielfs7, line 1535: I know the interface says it return string, but why we had casting before. Just double checking if it is really safe to remove :)

shpran: I will check it 😄

PR #1638, opened 2021-02-12
feature/CON-458/delete stale tmp folders after delivery exec

voknil:
deleting stale tmp folders after delivery exec


shpran (commented):
Good job 👍🏻 
Left some comments.

**Note: _we need to use correct [commit type](https://oat-sa.atlassian.net/wiki/spaces/DEV/pages/872775905/Conventional+Commits#Commit-types). Correct your commit message, please, if possible._**

PR #1649, opened 2021-03-01
fix:lib-tao-qti version

bartlomiejmarszal:
lib-tao-qti update

https://oat-sa.atlassian.net/browse/AUT-124


julien-sebire (approved):
LGTM

PR #1666, opened 2021-03-23
fix: Incorrect upperBound value saved within the Item

ampaveliev:
**Related to**
https://oat-sa.atlassian.net/browse/AUT-169

**Description**
Once the item author marks Max score range checkbox and then unchecks it, saves it, it adds "upper bound=0" to item XML that doesn't allow to get TT more than 0 scores for item with map response.

**Steps to reproduce**

1. Create a new item
2. Add choice interaction to the canvas
3. Go to Response and select "Map response" response processing
4. Enter scores > 0 for choices
5. Check Max Score range chack box and then uncheck it
6. Save
7. Go to Preview and select the choice with scores set > 0, submit


views/js/qtiCreator/widgets/helpers/formElement.js
```diff
@@ -282,7 +282,10 @@ define([
                     updateResponseDeclaration(element, value, options.updateCardinality);
                 }
 
-                if (!value && (element.is('orderInteraction') || element.is('graphicOrderInteraction'))) {
+                if ((this.name === 'upperBound') && this.disabled) {
+                    // if upperBound field triggered this callback and it is disabled, upperbound attribute should be removed.
+                    element[options.attrMethodNames.remove](name);
+                } else if (!value && (element.is('orderInteraction') || element.is('graphicOrderInteraction'))) {
```
bziondik, line 288: Maybe instead of add new condition we can update existing one?
You can ask UX why this is only for `orderInteraction` and `graphicOrderInteraction`
I think it should be removed if value is removed
```suggestion
                if (!value) {
```

bziondik (changes requested):
PR shouldn't contain bundles. Bundles will be created on release.

views/js/qtiCreator/widgets/helpers/formElement.js
habemuscode: https://oat-sa.atlassian.net/wiki/spaces/IMS/pages/1204585768/Max+min+values+in+QTI+interactions

We should modify `upperBound` in the same manner that max is modified

ampaveliev: > Maybe instead of add new condition we can update existing one? 
We can't because this function is called to handle other callback changes. For example, `maxChoice`

PR #1679, opened 2021-04-01
fix: TR-873 update background hover color for %selectableHover elements

lecosson:
Related to: https://oat-sa.atlassian.net/browse/TR-873

fix hover color for `%selectableHover`, make it same as highlighted by keyboard navigation

Requires:  none


views/scss/qti/_qti-base.scss
```diff
@@ -11,7 +11,7 @@ $mainContainer : '.qti-item';
 @import "inc/mediaelementplayer";
 
 %selectableHover {
-    background: whiten($info, .95) !important;
+    background: whiten($info, .9) !important;
```
jsconan, line 14: **nitpick:** This will introduce regressions as it will impact any theme.
The issue you try to fix is on a customer-specific theme. Please first lookt at it, and see how you can align the colors between the keyboard and mouse focus styles.

jsconan (changes requested):
Sorry, but I don't think the correct approach is applied there.

You try to fix an issue occurring on a custom theme by applying a change to every other theme. Unless the issue is spread over any themes, this is not the right approach. Instead, you should first try to see if the custom theme doesn't contain a discrepancy in the addressed style.

lecosson:
replaced by https://github.com/oat-sa/extension-tao-nfer/pull/290

PR #1680, opened 2021-04-01
Fix/aut 287/wrong behavoir when maxchoices not set

AnastasiaArcadia:
**Related to**
https://oat-sa.atlassian.net/browse/AUT-287

**Description**
QTI SDK was fixed: minChoices in HottextInteraction and PositionObjectInteraction is valid when greater than maxChoices if and only if the latter is 0.

**Steps to reproduce**
1. Create a new item
2. Add choice interaction to the canvas
3. Go to Response and select "Map response" response processing
4. Enter scores > 0 for choices
5. Check Max Score range chack box and then uncheck it
6. Save
7. Go to Preview and select the choice with scores set > 0, submit
8. SCORE and MAXSCORE should be displayed in console

Deployed on: http://test-anastasia.playground.kitchen.it.taocloud.org:44451


shpran (approved):
LGRM

bartlomiejmarszal (approved):
LGTM

PR #1754, opened 2021-05-31
Feature/AUT-36/Import & export items with rich passages

shpran:


model/Export/AbstractQTIItemExporter.php
```diff
@@ -316,4 +326,26 @@ protected function getMetadataExporter()
         }
         return $this->metadataExporter;
     }
+
+    private function copySharedStimulusCssFiles(string $link, string $basePath): string
+    {
+        $asset = new core_kernel_classes_Resource(UriHelper::decode($link));
```
mike-ionut-mihai-sandu-tao, line 332: i would break this into a different method and i would eliminate the static call if possible

model/Export/AbstractQTIItemExporter.php
```diff
@@ -316,4 +326,26 @@ protected function getMetadataExporter()
...
+        $baseName = 'assets/' . $this->getSharedStimulusCSSExporter()->getResourceLink($asset);
```
mike-ionut-mihai-sandu-tao, line 333: let's save `assets` into a constant and put directory_Separator or separate the '/'

model/Export/AbstractQTIItemExporter.php
```diff
@@ -316,4 +326,26 @@ protected function getMetadataExporter()
...
+        $this->getSharedStimulusCSSExporter()->pack(
+            $asset,
+            $this->getZip(),
+            $basePath . '/' . dirname($baseName)
```
mike-ionut-mihai-sandu-tao, line 337: because i saw that we are not compatible with Windows directory separator i think this is a `good way` for the moment so the hardcoded '/' is the way to go for now

model/qti/asset/AssetManager.php
```diff
@@ -129,6 +129,10 @@ public function importAuxiliaryFiles(QtiResource $qtiItemResource)
             $absolutePath = $this->getAbsolutePath($auxiliaryFile);
             $relativePath = $this->getRelativePath($qtiFile, $absolutePath);
 
+            if (is_dir($absolutePath) || $this->isAssetCssFolder($relativePath)) {
```
mike-ionut-mihai-sandu-tao, line 132: should we use the abstraction because we might have a different adapter ?

model/qti/asset/AssetManager.php
```diff
@@ -147,6 +151,13 @@ public function importAuxiliaryFiles(QtiResource $qtiItemResource)
         return $this;
     }
 
+    private function isAssetCssFolder(string $relativePath): bool
+    {
+        $relativePathParts = explode('/', dirname($relativePath));
```
mike-ionut-mihai-sandu-tao, line 156: maybe pathinfo will help you more on your check 
as if you have `./assets/' and you exploded that it will generate a different outcome then the expected one

model/Export/AbstractQTIItemExporter.php
bartlomiejmarszal: We should use TraitI would use 
`getResource` from `\oat\generis\model\OntologyAwareTrait`

model/Export/AbstractQTIItemExporter.php
```diff
@@ -316,4 +326,26 @@ protected function getMetadataExporter()
...
+        );
+
+        return $baseName;
+    }
+
+    private function getSharedStimulusCSSExporter(): SharedStimulusCSSExporter
+    {
+        if (!isset($this->sharedStimulusCSSExporter)) {
+            $this->sharedStimulusCSSExporter = $this->getServiceManager()->get(SharedStimulusCSSExporter::class);
```
bartlomiejmarszal, line 346: Why you using `getServiceManager` to retrieve service instead of `ServiceLocator`?

model/Export/AbstractQTIItemExporter.php
```diff
@@ -140,7 +143,14 @@ public function export($options = [])
                         $stream = $mediaSource->getFileStream($link);
                     }
 
-                    $baseName = ($mediaSource instanceof LocalItemSource) ? $link : 'assets/' . $mediaSource->getBaseName($link);
+                    $baseName = $link;
+
+                    if ($mediaSource instanceof MediaSource) {
+                        $baseName = $this->copySharedStimulusCssFiles($link, $basePath);
+                    } elseif (!$mediaSource instanceof LocalItemSource) {
+                        $baseName = 'assets/' . $mediaSource->getBaseName($link);
```
bartlomiejmarszal, line 151: Is this `/` directory separator?

model/Export/AbstractQTIItemExporter.php
```diff
@@ -140,7 +143,14 @@ public function export($options = [])
                         $stream = $mediaSource->getFileStream($link);
                     }
 
-                    $baseName = ($mediaSource instanceof LocalItemSource) ? $link : 'assets/' . $mediaSource->getBaseName($link);
+                    $baseName = $link;
+
+                    if ($mediaSource instanceof MediaSource) {
+                        $baseName = $this->copySharedStimulusCssFiles($link, $basePath);
+                    } elseif (!$mediaSource instanceof LocalItemSource) {
```
bartlomiejmarszal, line 150: What will happen when media source is nether `LocalItemSource` nor `MediaSource`?

bartlomiejmarszal (changes requested):
There is multiple business logic changes and no unit tests where modified. How come?

model/Export/AbstractQTIItemExporter.php
pnal: Related to this: https://github.com/oat-sa/extension-tao-mediamanager/pull/276/files#diff-dfe8daf8abc904a842e60b83b67c3aac0ccaa187cb4fc205c5bafc6fa07fab3cR162
Why you removed the $link param from `pack` method then added `getResourceLink` to `SharedStimulusCSSExporter` and again use it from here to specify zipPath but do not using initial resourceLink from $assets 3 lines above? Is it a different resource? I suggest to keep separate a new logic related to the Item export and adapt income data outside  `SharedStimulusCSSExporter` to keep it's responsibility clear and avoid breaking changes

pnal (commented):
Left some comments

model/Export/AbstractQTIItemExporter.php
bartlomiejmarszal: We need to leave ServiceManager call as the class is not ConfigurableService.

model/qti/asset/AssetManager.php
bartlomiejmarszal: pathinfo will not help much in this case. I'm worried that we are considering only single scenario where our path contain expected structure. I' not sure if this can be simplify this way. What if we have different structure
? Can we be sure that it will always looks similar to this:
```assets/9244e5cf60b621670d905/css/tao-user-styles.css```

File: model/qti/asset/handler/SharedStimulusAssetHandler.php [hunk 9]
```diff
@@ -94,7 +94,7 @@ public function handle($absolutePath, $relativePath)
 
         SharedStimulusImporter::isValidSharedStimulus($absolutePath);
         $newXmlFile = SharedStimulusPackageImporter::embedAssets($absolutePath);
```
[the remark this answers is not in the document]
bartlomiejmarszal: Yes it is. It's not inn scope of changes that were done to enable import/export css asset file.

PR #1855, opened 2021-08-19
Feature/TR-1810/Disable the media preview under Firefox 87+ when having a media interaction with WebM

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-1810

Requires: 
 - [ ] https://github.com/oat-sa/tao-item-runner-qti-fe/pull/211
 - [ ] https://github.com/oat-sa/tao-core-ui-fe/pull/346
 - [ ] https://github.com/oat-sa/tao-core/pull/3070
 - [ ] Once the companion PR is merged, update the package.json and the composer.json files accordingly

Disable the media preview when this is Firefox 87+ and the media interaction targets a WebM content.

**How to test:**
- Have a test containing media interactions, especially with video media. You need to have at least a video in WebM format. But other formats should also be checked as well.
- Check the behavior under various browsers: under Firefox, from version 87 on, if the media is WebM, the preview should be disabled, otherwise it should be presented

PR #1859, opened 2021-08-23
Backport/TR-1810/Disable media player preview under firefox

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-1810

Requires: 
 - [ ] https://github.com/oat-sa/tao-item-runner-qti-fe/pull/212
 - [ ] https://github.com/oat-sa/tao-core-ui-fe/pull/347
 - [ ] https://github.com/oat-sa/tao-core/pull/3077
 - [ ] Once the companion PR is merged, update the package.json and the composer.json files accordingly

Backport of #1855  to Sprint 134.5.100

Disable the media preview when this is Firefox 87+ and the media interaction targets a WebM content.

**How to test:**
- Have a test containing media interactions, especially with video media. You need to have at least a video in WebM format. But other formats should also be checked as well.
- Check the behavior under various browsers: under Firefox, from version 87 on, if the media is WebM, the preview should be disabled, otherwise it should be presented

### DIFF
--- a/manifest.php
+++ b/manifest.php
@@ -43,7 +43,7 @@ return [
     'name'        => 'taoQtiItem',
     'label'       => 'QTI item model',
     'license'     => 'GPL-2.0',
-    'version'     => '25.7.2.5',
+    'version'     => '25.7.2.5.1',
     'author'      => 'Open Assessment Technologies',
     'requires' => [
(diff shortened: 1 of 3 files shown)


jsconan:
Wrong base version

PR #1917, opened 2021-10-07
backport: AUT-1185 pcre limit handler

bartlomiejmarszal:
- Backport: [AUT-1185](https://oat-sa.atlassian.net/browse/AUT-1185) pcre limit handler #1913

### DIFF
--- a/model/qti/container/Container.php
+++ b/model/qti/container/Container.php
@@ -22,6 +22,7 @@
 
 namespace oat\taoQtiItem\model\qti\container;
 
+use Monolog\Logger;
 use oat\taoQtiItem\model\qti\Element;
 use oat\taoQtiItem\model\qti\IdentifiedElementContainer;
 use oat\taoQtiItem\model\qti\Item;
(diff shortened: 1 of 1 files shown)


model/qti/container/Container.php
```diff
@@ -222,6 +223,19 @@ public function fixNonvoidTags($html)
             // correctly closed element
             return trim(mb_substr($matches[0], 0, -2), 'UTF-8') . '></' . $matches[2] . '>';
         }, $html);
+
+        $pregLastError = preg_last_error();
```
andreluizmachado, line 227: Why did you remove the private method?

model/qti/container/Container.php
```diff
@@ -22,6 +22,7 @@
 
 namespace oat\taoQtiItem\model\qti\container;
 
+use Monolog\Logger;
```
pnal, line 25: did not notice that in the initial PR, 
asking now, what is the purpose of adding it here?

model/qti/container/Container.php
bartlomiejmarszal: Extracting this method with argument type hint cause issues because `$content` can be string, array or null. Extracting this business logic to separate method was useless and less readable.

andreluizmachado (approved):

PR #1923, opened 2021-10-15
fix/TR-1959/graphic-associate-interaction-max-association-issue

marpesia:
**Related to:**

- [TR-1959](https://oat-sa.atlassian.net/browse/TR-1959)

**Description:**
The response is not saving on a single association because the output is not a pair. The code is waiting for a pair response.

**Changes:**

- Creating an exception and adding the single response as a pair to the object solves the issue.

**How to check:**

- There is a min_max_test.zip on the ticket with some different cases.
- In the ticket, there is a diagram with cases that work and don't. It used to fail when maxAssociations in 1, no matter if minAssociations is set or not.

**Requires:**
- [TaoCommunity composer.json](https://github.com/oat-sa/tao-community/blob/2021.10/composer.json)
- Install package linked [oat-sa/tao-item-runner-qti-fe](hotfix/TR-1959/graphic-associate-interaction-max-association-issue)


krampstudio (approved):
It works as expected, the response is restored! Well done.

PR #1975, opened 2021-12-08
TR-2961 text reader asset extraction

poyuki:
# [TR-2961](https://oat-sa.atlassian.net/browse/TR-2961)

## Summary
Need to create functionality which allow publish inline base64 encoded PCI Text Reader item assets to cdn
 
## How to test
* update extension by branch
* create `item` with `Text Reader` with image
* create `test` with created in prev step `item`
* publish this `test` as local
* run published delivery and check in browser inspector image `src`

For affecting local `env` without `cdn` need to update `taoQtiItem/model/compile/QtiAssetReplacer/NullQtiItemAssetReplacer.php`
by next code 
```
class NullQtiItemAssetReplacer extends ConfigurableService implements QtiItemAssetReplacer
{
    public function shouldBeReplaced(PackedAsset $packetAsset): bool
    {
        return $packetAsset->getMediaAsset()->getMediaSource() instanceof InlineSource;
    }

    public function replace(PackedAsset $packetAsset, $itemId): PackedAsset
    {
        $mediaAsset = $packetAsset->getMediaAsset();
        $filename = $mediaAsset->getMediaSource()->getBaseName($mediaAsset->getMediaIdentifier());

        return $packetAsset->setReplacedBy($filename);
    }
}
```


model/qti/CustomInteractionAsset/DataUrlMimeTypeDecoder.php
```diff
@@ -0,0 +1,111 @@
...
+     * @param string $dataUrl
+     * @return string|null
+     * @throws NotDataUrlProvidedException
+     */
+    public static function decodeToAssetType(string $dataUrl): ?string
+    {
+        if (self::checkIsDataUrl($dataUrl)) {
+            $mimeType = mime_content_type($dataUrl);
+
+            return self::mapMimeTypeToAsset($mimeType);
```
wazelin, line 94: `$mimeType` can potentially be a `false`.

poyuki: fixed

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
```diff
@@ -0,0 +1,54 @@
...
+use oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor\BaseExtendedCustomInteractionAssetExtractor;
+use oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor\TextReaderExtendedAssetExtractor;
+use oat\taoQtiItem\model\qti\interaction\CustomInteraction;
+use oat\taoQtiItem\model\qti\interaction\ImsPortableCustomInteraction;
+use oat\taoQtiItem\model\qti\interaction\PortableCustomInteraction;
+
+/**
+ * @author Kiryl Poyu <kyril.poyu@taotesting.com>
+ */
+class ExtendedCustomInteractionAssetExtractorAllocator
```
wazelin, line 36: This looks more like an abstract factory to me, WDYT?

I mean, it clearly _creates_ an extractor service.

poyuki: I do not think so. In general, this class is used just to determine which implementation of the base class we need. An abstract factory can be used as part of this flow if we have a complex process for creating descendants.

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
```diff
@@ -0,0 +1,54 @@
...
+        if (
+            $interaction instanceof ImsPortableCustomInteraction
+            || $interaction instanceof PortableCustomInteraction
+        ) {
+            if ($interaction->getTypeIdentifier() === self::TEXT_READER_INTERACTION) {
+                return new TextReaderExtendedAssetExtractor($interaction);
+            }
+        }
+
+        return null;
```
wazelin, line 52: Instead of a `null`. WDYT of returning a no-op implementation of the interface?

poyuki: probably we can, but i don't see any real profit from this. We just remove 1 `if` but create empty class which should return empty array for this iteration (IMO it looks strange)
```php
if ($assetExtractor !== null) {
            $extractedAssets = $assetExtractor->extract();
            foreach ($extractedAssets as $asset) {
                $this->addAsset('pci-resource', $asset);
            }
        }
```

wazelin: That's the point of a no-op service. Yes, you have one `if` now, but you may as well get some other at some other place.
null-objects instead of `null` values allow reducing the complexity of the code.

I won't be pushing for that change any more, but suggest you think about it a bit :)

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
```diff
@@ -0,0 +1,54 @@
...
+/**
+ * @author Kiryl Poyu <kyril.poyu@taotesting.com>
+ */
+class ExtendedCustomInteractionAssetExtractorAllocator
+{
+    public const TEXT_READER_INTERACTION = 'textReaderInteraction';
+
+    public static function allocateExtractor(
+        CustomInteraction $interaction
+    ): ?BaseExtendedCustomInteractionAssetExtractor {
```
wazelin, line 42: I'd prefer if this method returned an interface.
An abstract class is good to have a common piece of logic. A return type should first of define a public contract, which is the main purpose of the interfaces.

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
```diff
@@ -0,0 +1,56 @@
...
+ */
+
+namespace oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor;
+
+use oat\taoQtiItem\model\qti\interaction\ImsPortableCustomInteraction;
+use oat\taoQtiItem\model\qti\interaction\PortableCustomInteraction;
+
+/**
+ * @author Kiryl Poyu <kyril.poyu@taotesting.com>
+ */
```
wazelin, line 32: I think you can put that to the header above, right under the copyright.

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
```diff
@@ -0,0 +1,56 @@
+<?php
+
+declare(strict_types=1);
```
wazelin, line 3: This is the order of blocks as suggested by PSR-12

1. Opening <?php tag.
2. File-level docblock.
3. One or more declare statements.
4. ...

`declare()` therefore must be _under_ the file-level license block.

model/qti/AssetParser.php
```diff
@@ -356,6 +359,17 @@ private function getXmlProperties($properties)
         return $xmls;
     }
 
+    private function extractAdvancedCustomInteractionAssets(CustomInteraction $interaction): void
+    {
+        $assetExtractor = ExtendedCustomInteractionAssetExtractorAllocator::allocateExtractor($interaction);
+        if ($assetExtractor !== null) {
```
wazelin, line 365: If you decide not to implement a no-op service here, could you please consider adding an early `return` instead of a logic with an `if` block?
That improves readability, keeping the indentation not too nested.

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
poyuki: I agree, in case of repeated use there will be more `if`. You pushed it through :)

model/qti/AssetParser.php
poyuki: i will create no-op service

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
poyuki: agree but template here confused me a little bit 
https://oat-sa.atlassian.net/wiki/spaces/~369998401/pages/186221874/PhpStorm+live+templates

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
poyuki: agree, make sense

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
poyuki: LGTM, you just need to understand where this mapping will come from

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
```diff
@@ -0,0 +1,54 @@
...
+        if (
+            $interaction instanceof ImsPortableCustomInteraction
+            || $interaction instanceof PortableCustomInteraction
+        ) {
+            if ($interaction->getTypeIdentifier() === self::TEXT_READER_INTERACTION) {
```
wazelin, line 47: Let's make `getTypeIdentifier()` part of `CustomInteraction`.

In order to keep backwards compatibility, it can simply return an empty `string` by default, while the descendants implement the real deal.

One more `if` gone :)

wazelin (changes requested):
Overall, the solution seems to be clear and extendable.
I do have a number of nitpicks, though :)

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
wazelin: Welp, you spot an issue with it :)

wazelin: Here's the common doc BTW
https://oat-sa.atlassian.net/wiki/spaces/NEX/pages/87361761/How+To+Insert+Open+Assessment+Technologies+SA+Licence+Header+Comment

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
poyuki: In some other classes this used as class dockblock

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
wazelin: I suppose we can do that on the service provider level.
https://github.com/oat-sa/generis/blob/master/core/DependencyInjection/ContainerServiceProviderInterface.php

poyuki: probably in this case extractorMap should contain factories set  with some interface
```php
interface CustomInteractioAssetInstractorFactoryInterface {
   public function create(CustomInteraction $interaction): AssetExtractorInterface
}
```
and allocator in this case will be look like 
```php
public function __constuct(private array $extractorMap) {
}

public function allocateExtractor(CustomInteraction $interaction) {
    return $this->extractorMap[$interaction->getTypeIdentifier()]->create(interaction) ?? new NullAssetExtractor();
}
```
to keep flexibility in creation of some complex extractors

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
wazelin: I don't think it should be that way to be honest.
We usually use a file-level docblock and that's it.

model/qti/CustomInteractionAsset/Extractor/Api/AssetExtractorInterface.php
```diff
@@ -0,0 +1,36 @@
...
+
+/**
+ * @author Kiryl Poyu <kyril.poyu@taotesting.com>
+ */
+interface AssetExtractorInterface
+{
+    /**
+     * @return array<string>
+     */
+    public function extract(CustomInteraction $interaction): array;
```
wazelin, line 35: Just a suggestion, WDYT of returning an `iterable` type here?
That will allow you to replace the implementation with a generator, instead of building the whole array of assets.

model/qti/CustomInteractionAsset/Extractor/TextReaderAssetExtractor.php
```diff
@@ -0,0 +1,63 @@
...
+    public const CONTENT_PREFIX = 'content-';
+
+    /**
+     * @inheritDoc
+     */
+    public function extract(CustomInteraction $interaction): array
+    {
+        $extractedAssets = [];
+        $contentProperties = array_filter($interaction->getProperties(), static function ($propertyKey) {
+            return strpos($propertyKey, self::CONTENT_PREFIX) !== false;
```
wazelin, line 45: Considering the `'content-'` value is a prefix, WDYT of checking if the key actually starts from it, not just contains it?
```suggestion
            return strpos($propertyKey, self::CONTENT_PREFIX) === 0;
```

wazelin (changes requested):

model/qti/CustomInteractionAsset/ExtendedCustomInteractionAssetExtractorAllocator.php
wazelin: Or that map itself can be a collection of callbacks, instantiating a required extractor.

model/qti/CustomInteractionAsset/Extractor/BaseExtendedCustomInteractionAssetExtractor.php
poyuki: i think you are right, but i started do that in following this doc and other class docblock
https://oat-sa.atlassian.net/wiki/spaces/OAT/pages/4161651/PHP

test/unit/model/qti/CustomInteractionAsset/CustomInteractionAssetExtractorAllocatorTest.php
```diff
@@ -0,0 +1,64 @@
...
+ * Copyright (c) 2021 (original work) Open Assessment Technologies SA ;
+ */
+
+namespace oat\taoQtiItem\test\unit\model\qti\CustomInteractionAsset;
+
+use oat\generis\test\TestCase;
+use oat\taoQtiItem\model\qti\CustomInteractionAsset\CustomInteractionAssetExtractorAllocator;
+use oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor\NullAssetExtractor;
+use oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor\TextReaderAssetExtractor;
+use Ramsey\Uuid\Uuid;
```
wazelin, line 29: I don't think this particular extension requires that lib by default.
Please either
1. Require it as a `dev` dependency
2. Use something like `uniqid()` instead, which should be sufficient for the tests.

poyuki: done

model/qti/CustomInteractionAsset/Extractor/TextReaderAssetExtractor.php
poyuki: LGTM

model/qti/CustomInteractionAsset/Extractor/Api/AssetExtractorInterface.php
poyuki: agree with you

wazelin (changes requested):
Those are not your changes, but the tests are failing on that extension. Let's fix them!

```
PHPUnit 8.5.21 by Sebastian Bergmann and contributors.


Failed asserting that two arrays are identical.
<Click to see difference>

 /Users/sergei.mikhailov/Sources/nextgen-stack/tao/taoQtiItem/test/unit/model/QtiJsonItemCompilerTest.php:169
 

TypeError : Argument 1 passed to oat\tao\model\media\MediaAsset::__construct() must implement interface oat\tao\model\media\MediaBrowser, null given, called in /Users/sergei.mikhailov/Sources/nextgen-stack/tao/taoItems/models/classes/media/ItemMediaResolver.php on line 45
 /Users/sergei.mikhailov/Sources/nextgen-stack/tao/tao/models/classes/media/MediaAsset.php:45
 /Users/sergei.mikhailov/Sources/nextgen-stack/tao/taoItems/models/classes/media/ItemMediaResolver.php:45
 /Users/sergei.mikhailov/Sources/nextgen-stack/tao/taoQtiItem/model/compile/QtiAssetCompiler/QtiItemAssetCompiler.php:109
 /Users/sergei.mikhailov/Sources/nextgen-stack/tao/taoQtiItem/model/compile/QtiAssetCompiler/QtiItemAssetCompiler.php:68
 /Users/sergei.mikhailov/Sources/nextgen-stack/tao/taoQtiItem/test/unit/model/compile/QtiItemAssetCompilerTest.php:299
```

I've just run that under PHP 7.4.

test/unit/model/qti/CustomInteractionAsset/CustomInteractionAssetExtractorAllocatorTest.php
```diff
@@ -0,0 +1,65 @@
...
+     */
+    private $subject;
+
+    protected function setUp(): void
+    {
+        parent::setUp();
+        $this->subject = new CustomInteractionAssetExtractorAllocator([
+            TextReaderAssetExtractor::INTERACTION_IDENTIFIER => static function () {
+                return new TextReaderAssetExtractor();
+            }
```
wazelin, line 46:
```suggestion
            TextReaderAssetExtractor::INTERACTION_IDENTIFIER => new TextReaderAssetExtractor(),
```

wazelin (changes requested):

wazelin (approved):

model/qti/AssetParser.php
```diff
@@ -356,6 +360,16 @@ private function getXmlProperties($properties)
         return $xmls;
     }
 
+    private function extractAdvancedCustomInteractionAssets(CustomInteraction $interaction): void
+    {
+        $extractorAllocator = $this->getCustomInteractionAssetExtractorAllocator();
+        $extractor = $extractorAllocator->allocateExtractor($interaction->getTypeIdentifier());
+
+        foreach ($extractor->extract($interaction) as $asset) {
+            $this->addAsset('apip', $asset);
```
augustas, line 369: Why all these extractors would add assets as `apip`? `apip` is a specific type of accessibility related stuff http://www.imsglobal.org/apip/index.html

composer.json
```diff
@@ -61,6 +61,8 @@
     },
     "minimum-stability": "dev",
     "require": {
+        "ext-fileinfo": "*",
+        "ext-simplexml": "*",
```
augustas, line 65: `ext-simplexml` is required by `qti-sdk`, [here](https://github.com/oat-sa/qti-sdk/blob/master/composer.json#L33), and this repository wouldn't work without `qti-sdk`.
Why did you need to add `ext-fileinfo` requirement?

augustas (changes requested):
Great work 👍🏻  As I understand the flexibility here is to have the possibility to add asset extractors for other PCIs later, not only text reader interaction. 
Why do we need to limit the data URLs parsing only to text reader interaction? Why not make all PCI data URLs extracted? Are there custom non html tags that have to be parsed?
Could `AssetParser::loadCustomElementAssets(Element $element)` be refactored to extract the xml xpath asset parsing to the newly introduced `AssetExtractor` abstraction?

composer.json
poyuki: previously i used `mime_content_type` for for determining asset type. but after this behavior was simplified. And now `ext-fileinfo` can be deleted

model/qti/AssetParser.php
poyuki: i was trying to define type of asset by mime-type from data url, but it was increased complexity of the component by supporting mapping mime types to assets types. and i decide to use something common, cause we just need to push in assets result array and we have restriction of supported assets type. Do you see any critical problems in this case? maybe you can advise something for that usage?

poyuki:
@augustas 
> Why do we need to limit the data URLs parsing only to text reader interaction? Why not make all PCI data URLs extracted? 

I was solving a specific problem for text reader. Probably this abstraction can be extended and improved, but for me not clear all possible aspect right now.

> Are there custom non html tags that have to be parsed?

We can cover this case by extending this component or creating additional.

> Could AssetParser::loadCustomElementAssets(Element $element) be refactored to extract the xml xpath asset parsing to the newly introduced AssetExtractor abstraction?

Probably we can do it as a tech dep task in future

yohanlaborda (changes requested):

test/unit/model/qti/CustomInteractionAsset/Extractor/TextReaderExtendedAssetExtractorTest.php
```diff
@@ -0,0 +1,118 @@
...
+     */
+    private function generateCustomInteractionWithProperties(): CustomInteraction
+    {
+        $properties = [];
+        for ($i = 0, $maxAssets = random_int($i, 10); $i < $maxAssets; $i++) {
+            $dataUrl = "data:image/jpeg;base64," . uniqid('test', true);
+            $properties[TextReaderAssetExtractor::CONTENT_PREFIX . uniqid('test', true)] = $dataUrl;
+        }
+
+        return new class($properties) extends CustomInteraction {
```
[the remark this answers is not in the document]
poyuki: probably, but i think it's not necessary, by usage anonymous class i can emulate needed only for exact test behavior

model/qti/CustomInteractionAsset/Extractor/TextReaderAssetExtractor.php
```diff
@@ -0,0 +1,55 @@
...
+
+use oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor\Api\AssetExtractorInterface;
+use oat\taoQtiItem\model\qti\interaction\CustomInteraction;
+
+/**
+ * @author Kiryl Poyu <kyril.poyu@taotesting.com>
+ */
+class TextReaderAssetExtractor implements AssetExtractorInterface
+{
+    public const INTERACTION_IDENTIFIER = 'textReaderInteraction';
```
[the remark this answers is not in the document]
poyuki: can you add more details to your comment?

model/qti/CustomInteractionAsset/Extractor/NullAssetExtractor.php
```diff
@@ -0,0 +1,39 @@
...
+declare(strict_types=1);
+
+namespace oat\taoQtiItem\model\qti\CustomInteractionAsset\Extractor;
+
+use oat\taoQtiItem\model\qti\interaction\CustomInteraction;
+
+/**
+ * @author Kiryl Poyu <kyril.poyu@taotesting.com>
+ */
+class NullAssetExtractor implements Api\AssetExtractorInterface
```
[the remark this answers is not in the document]
poyuki: good point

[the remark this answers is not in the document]
wazelin: https://en.wikipedia.org/wiki/Null_object_pattern

wazelin:
@augustas, since you're appearing in this PR (thank you for that BTW!), I'm gonna use it to send a message to you.

Please ping in Slack _today_ when you have a minute.

model/qti/CustomInteractionAsset/Extractor/NullAssetExtractor.php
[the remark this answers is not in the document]
poyuki: renaming the extractor to `empty` just make it semantically closer to behavior which class provide, WDYT @wazelin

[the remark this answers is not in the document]
wazelin: IMO that's still a `Null` object. Let's keep it as is in order to correspond to the design pattern name.

model/qti/CustomInteractionAsset/Extractor/TextReaderAssetExtractor.php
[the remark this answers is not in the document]
yohanlaborda: When you put a public constant, anyone can use it in another class or part of the code, that creates dependency on your code and if you change that constant it can break your code.

If you need to use it in your test, you only need to copy the value in the test, never put the reference to the constant. If someone changes the code, the test should break.

If you need the value to use it in various parts of the code, you could create an Enum (for now only available in version 8.1) or create a ValueObject that would behave like an Enum.

test/unit/model/qti/CustomInteractionAsset/Extractor/TextReaderExtendedAssetExtractorTest.php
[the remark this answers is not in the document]
yohanlaborda: When you break into class, you can put a more explicit name and reuse the class in a different test.

This is useful to clarify the code, but it is not necessary in all cases. For me it is better to create many classes with an explicit name.

model/qti/CustomInteractionAsset/Extractor/TextReaderAssetExtractor.php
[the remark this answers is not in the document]
poyuki: I understand a concept. But i don't think it always necessary increase complexity of component in this way. it's fair for `CONTENT_PREFIX` which created in general for internal usage.

yohanlaborda (approved):

model/qti/AssetParser.php
augustas: Well, you could use a [map from elsewhere](https://github.com/thephpleague/mime-type-detection/blob/main/src/GeneratedExtensionToMimeTypeMap.php), but I agree there is no great benefit to learn the exact asset type here. Yet `apip` is a bit misleading. You used term `advanced` in the method name to differentiate theses extracted assets, maybe it could be used as a type here as well

poyuki: usage of some new type there create a need to extend this map https://github.com/oat-sa/extension-tao-item/blob/master/models/classes/pack/ItemPack.php#L48

augustas: Okay, but it doesn't change the fact that `apip` type here is incorrect and it's a shortcut. If no time to refactor, by returning a proper type or creating a new type, I think it would be better to add a comment why `apip` type was used here

augustas (approved):

PR #2001, opened 2022-01-19
Fix/tr 3390 regenerate arabic translation

btamas:
Related to https://oat-sa.atlassian.net/browse/TR-3390

Regenerate translations, because of https://github.com/oat-sa/tao-core/pull/3342


oatymart (approved):
Ctrl+F shows that math symbols were removed as intended.

svoronuk (approved):
LGFM

PR #2032, opened 2022-02-17
Feature/rel 612/fix jp translations

siwane:
**Description**
Updates to allow translations + include Japanese translations

**Ticket**
https://oat-sa.atlassian.net/browse/INF-87


alroniks:
@siwane Why additional Japanese languages are removed? Is it approved by the customer?

PR #2050, opened 2022-03-06
fix: Wrong qti version during package export

yaraslau-kavaliou:
https://oat-sa.atlassian.net/browse/OATSD-1707

I can assume that the reason of the issue was switching from v2p1 to v2p2 version in qti.xml package itself which was not reflected in the export functionality.

### DIFF
--- a/model/Export/AbstractQTIItemExporter.php
+++ b/model/Export/AbstractQTIItemExporter.php
@@ -210,9 +210,7 @@ abstract class AbstractQTIItemExporter extends taoItems_models_classes_ItemExpor
             $report->setType(\common_report_Report::TYPE_ERROR);
         }
 
-        // Set the right qti version
-        $content = preg_replace('/(http:\/\/www\.imsglobal\.org.+?)qtiv(\w\w\w)/', '$1qtiv' . $this->getQTIVersion(), $content);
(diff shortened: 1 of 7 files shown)


model/Export/AbstractQTIItemExporter.php
```diff
@@ -208,6 +210,10 @@ public function export($options = [])
             $report->setType(\common_report_Report::TYPE_ERROR);
         }
 
+        // Set the right qti version
+        $content = preg_replace('/(http:\/\/www\.imsglobal\.org.+?)qtiv(\w\w\w)/', '$1qtiv' . $this->getQTIVersion(), $content);
```
julien-sebire, line 214: I'm afraid this might replace strings we don't want to replace. Can we be a bit more specific here?
Unit tests would be great.
`+?`: I don't understand the `?` after the `+` 
`(\w\w\w)`: the second `\w` can be replaced by a `p`

julien-sebire (changes requested):
Unit tests + regex to make more precise.

model/Export/AbstractQTIItemExporter.php
yaraslau-kavaliou: Made regexp a bit more precise.
Add unit test.
+? - means lazy quantifier

julien-sebire (approved):
All good. Thanks Yaraslau!

tikhanovichA (approved):

PR #2114, opened 2022-04-28
fix: reset the previous definition of the XML, to receive updated passage

bziondik:
Related to: https://oat-sa.atlassian.net/browse/AUT-2028
**Depends on:**
- [ ] https://github.com/oat-sa/tao-item-runner-qti-fe/pull/273

**Problem:** XML is loaded only once for Items page
**Solution:** reset the previous definition of the XML, to receive updated passage

## How to test:

- Create a item with shared stimulus
- Open new tab with media, editing the shared stimuli, save
- Going back to the tab with item, closing the item by "mange items"
- Then clicking "authoring again"

**Actual:**  The shared stimuli change is not included
**Expected:** The item is updated with the new change in shared stimuli.


Silvia-loza (approved):
Great the added  comment really helps making clear how this works.
Just package to be updated when released,  all LGTM.

PR #2119, opened 2022-05-03
update tao-core dependency to 50.10.0

shaveko:
related to https://oat-sa.atlassian.net/browse/AUT-1980

### Description 

extension uses functionality introduced in tao-core 50.10.0 (services/features) `isVisible` feature visibility utility function

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,7 +66,7 @@
         "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
         "naneau/semver": "~0.0.7",
         "oat-sa/generis" : ">=14.4.0",
-        "oat-sa/tao-core" : ">=49.3.0",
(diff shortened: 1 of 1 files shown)


krampstudio (commented):
This looks good except that it should be a feature version update, because of the jump of the dependency between 49.x.x and 50.x.x
Can you amend the commit otherwise we can squash the PR on merge

shaveko:
> This looks good except that it should be a feature version update, because of the jump of the dependency between 49.x.x and 50.x.x Can you amend the commit otherwise we can squash the PR on merge

I will close this PR, dependency version bump will go in PR vs [`integration/AUT-1979/hide-unsupported-features` ](https://github.com/oat-sa/extension-tao-itemqti/tree/integration/AUT-1979/hide-unsupported-features)

PR #2128, opened 2022-05-06
Fix/TR-3995/Max words restriction can be exceeded

jsconan:
4/4

Related to: https://oat-sa.atlassian.net/browse/TR-3995

Requires: 
 - [x] https://github.com/oat-sa/tao-item-runner-qti-fe/pull/283
 - [x] update the `package.json` file once the companion PR has been merged

The extended text interaction was allowing typing more words than the allowed constraint.
This was due to some wrong conditions in the keystroke handler and a missing typecast when reading the constraint properties.

How to test:
- create a test having extended text interactions with a constraint for the max words and for the max length (a test sample is attached to the ticket).
- publish the test, then take it
- type some text in the extended text interaction, then copy it and paste it until the limit is reached.
- try to exhaust the limit by typing more words
- add a new line by pressing the enter key
- try again to exhaust the limit by typing more words

Without the fix, the limit can be overflown.
With the fix, we cannot overflow the limit.


meefox (approved):

AntonNeviarouski:
QA verified

PR #2155, opened 2022-06-16
fix: Prevent value overwrite for those already set on item import

hectoras:
**Associated Jira issue:** [ADF-1073](https://oat-sa.atlassian.net/browse/ADF-1073)

Previously, `GenericLomOntologyClassificationInjector` was calling `editPropertyValueByLg()` (which overwrites all values for a property) or `setPropertyValueByLg()` (which appends them) depending on the number of _previous_ values already set for that property. However, those previous values may have been set by this very method itself in previous iterations.

```
if ($target->getPropertyValuesByLg($prop, $lang)->count() > 0) {
    $target->editPropertyValueByLg($prop, $metadataValue->getValue(), $lang); // overwrite
} else {
    $target->setPropertyValueByLg($prop, $metadataValue->getValue(), $lang); // appends
}
```

This PR changes that to set all values at once (and only in case the property definition allows it, i.e. if `isMultiple()`).

```
if ($numPreviousValues > 0 && $propertyInstance->isMultiple()) {
    $target->setPropertyValueByLg($propertyInstance, $value, $langCode); // append
} else {
    $target->editPropertyValueByLg($propertyInstance, $value, $langCode);
}
```

Additionally, this fixes some minor formatting issues raised by the PSR-12 linter ~~and adds (keeps) additional calls to the logger used to debug the problem in the first place~~.

---

**TODO**

- [x] Check unit tests are still passing
- [x] Add additional unit tests

### DIFF
--- a/model/qti/metadata/AbstractMetadataService.php
+++ b/model/qti/metadata/AbstractMetadataService.php
@@ -15,15 +15,13 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017-2022 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
  *
  */
 
 namespace oat\taoQtiItem\model\qti\metadata;
 
-use common_Logger;
 use oat\oatbox\service\ConfigurableService;
-use Psr\Log\LoggerInterface;
 
 /**
  * Class AbstractMetadataService
@@ -55,20 +53,6 @@ abstract class AbstractMetadataService extends ConfigurableService
     protected $instances;
 
     /**
-     * @var LoggerInterface
-     */
-    private $logger;
-
-    public function __construct($options = [], LoggerInterface $logger = null)
-    {
-        parent::__construct($options);
-
-        $this->logger = ($logger === null
-            ? common_Logger::singleton()->getLogger()
-            : $logger);
-    }
-
-    /**
      * Allow to register into config a metadataService
      */
     abstract protected function registerMetadataService();
@@ -88,10 +72,7 @@ abstract class AbstractMetadataService extends ConfigurableService
             $metadata[] = $extractor->extract($source);
         }
         $metadata = array_merge_recursive(...$metadata);
-
-        $this->logger->debug(
-            __('%s metadata values found in source by extractor(s).', count($metadata))
-        );
+        \common_Logger::d(__('%s metadata values found in source by extractor(s).', count($metadata)));
 
         return $metadata;
     }
@@ -101,26 +82,18 @@ abstract class AbstractMetadataService extends ConfigurableService
      *
      * Inject metadata value for an identifier by calling each injectors
      * Injectors need $target for injection
+     *
+     * @param $identifier
+     * @param $target
      */
     public function inject($identifier, $target)
     {
         if ($this->hasMetadataValue($identifier)) {
-            $this->logger->info(
-                __('Preparing Metadata Values for target "%s"...', $identifier)
-            );
-
+            \common_Logger::i(__('Preparing Metadata Values for target "%s"...', $identifier));
             $values = $this->getMetadataValue($identifier);
 
             foreach ($this->getInjectors() as $injector) {
-                $this->logger->info(
-                    __(
-                        'Attempting to inject "%s" metadata values for target "%s" with metadata Injector "%s".',
-                        count($values),
-                        $identifier,
-                        get_class($injector)
-                    )
-                );
-
+                \common_Logger::i(__('Attempting to inject "%s" metadata values for target "%s" with metadata Injector "%s".', count($values), $identifier, get_class($injector)));
                 $injector->inject($target, [$identifier => $values]);
             }
         }
@@ -190,6 +163,7 @@ abstract class AbstractMetadataService extends ConfigurableService
                 $this->unregisterInstance(self::INJECTOR_KEY, $name);
                 break;
         }
+        return;
     }
 
     /**
--- a/model/qti/metadata/AbstractMetadataService.php
+++ b/model/qti/metadata/AbstractMetadataService.php
@@ -54,7 +54,9 @@ abstract class AbstractMetadataService extends ConfigurableService
      */
     protected $instances;
 
-    /** @var LoggerInterface */
+    /**
+     * @var LoggerInterface
+     */
     private $logger;
 
     public function __construct($options = [], LoggerInterface $logger = null)
--- a/model/qti/metadata/AbstractMetadataService.php
+++ b/model/qti/metadata/AbstractMetadataService.php
@@ -15,13 +15,15 @@
  * along with this program; if not, write to the Free Software
  * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301, USA.
  *
- * Copyright (c) 2017 (original work) Open Assessment Technologies SA;
+ * Copyright (c) 2017-2022 (original work) Open Assessment Technologies SA;
  *
  */
 
 namespace oat\taoQtiItem\model\qti\metadata;
 
+use common_Logger;
 use oat\oatbox\service\ConfigurableService;
+use Psr\Log\LoggerInterface;
 
 /**
  * Class AbstractMetadataService
@@ -53,6 +55,20 @@ abstract class AbstractMetadataService extends ConfigurableService
     protected $instances;
 
     /**
+     * @var LoggerInterface
+     */
+    private $logger;
+
+    public function __construct($options = [], LoggerInterface $logger = null)
+    {
+        $this->setOptions($options);
+
+        $this->logger = ($logger === null
+            ? common_Logger::singleton()->getLogger()
+            : $logger);
+    }
+
+    /**
      * Allow to register into config a metadataService
      */
     abstract protected function registerMetadataService();
@@ -72,7 +88,10 @@ abstract class AbstractMetadataService extends ConfigurableService
             $metadata[] = $extractor->extract($source);
         }
         $metadata = array_merge_recursive(...$metadata);
-        \common_Logger::d(__('%s metadata values found in source by extractor(s).', count($metadata)));
+
+        $this->logger->debug(
+            __('%s metadata values found in source by extractor(s).', count($metadata))
+        );
 
         return $metadata;
     }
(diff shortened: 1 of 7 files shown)


model/qti/metadata/AbstractMetadataService.php
```diff
@@ -52,6 +54,20 @@ abstract class AbstractMetadataService extends ConfigurableService
      */
     protected $instances;
 
+    /**
+     * @var LoggerInterface
+     */
+    private $logger;
```
gabrielfs7, line 60:
```suggestion
    /** @var LoggerInterface */
    private $logger;
```

model/qti/metadata/AbstractMetadataService.php
```diff
@@ -52,6 +54,20 @@ abstract class AbstractMetadataService extends ConfigurableService
...
+
+    public function __construct($options = [], LoggerInterface $logger = null)
+    {
+        parent::__construct($options);
+
+        $this->logger = ($logger === null
+            ? common_Logger::singleton()->getLogger()
+            : $logger);
```
gabrielfs7, line 68: We can get log from '$this->getServiceManager()->get(LoggerService::SERVICE_ID)', no? 

Why do we need to override the constructor?

model/qti/metadata/ontology/GenericLomOntologyClassificationInjector.php
```diff
@@ -25,11 +24,23 @@
...
 class GenericLomOntologyClassificationInjector implements MetadataInjector
 {
     use OntologyAwareTrait;
 
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct(LoggerInterface $logger = null)
+    {
+        $this->logger = ($logger ?? common_Logger::singleton()->getLogger());
```
gabrielfs7, line 41: Is there a way to avoid this static call?

model/qti/metadata/ontology/OntologyMetadataInjector.php
```diff
@@ -26,53 +25,67 @@
...
+
+    /** @var LoggerInterface */
+    private $logger;
+
+    public function __construct(
+        LoggerInterface $logger = null,
+        EventManager $eventManager = null
+    ) {
+        $this->eventManager = $eventManager;
+        $this->logger = ($logger ?? common_Logger::singleton()->getLogger());
```
gabrielfs7, line 49: No way to avoid static call?

model/qti/metadata/ontology/OntologyMetadataInjector.php
```diff
@@ -99,57 +112,116 @@ public function inject($target, array $values)
                     }
 
                     $data[$rule[0]][$lang][] = [$metadataValue->getValue(), $metadataValue];
+                } else {
+                    $this->debug(
```
gabrielfs7, line 116: Do not forget to remove the logs if not used

model/qti/metadata/ontology/OntologyMetadataInjector.php
```diff
@@ -99,57 +112,116 @@ public function inject($target, array $values)
...
+
         return false;
     }
+
+    private function debug(string $message, ...$replacements): void
+    {
+        $this->logger->info(__CLASS__ . ': ' . vsprintf($message, $replacements));
+    }
+
+    private function getEventManager(): EventManager
```
gabrielfs7, line 205: Why not do the same with the logs?

model/qti/metadata/AbstractMetadataService.php
hectoras: This domain service does still use `getOption()` in some places and there are at least two classes that inherit from it, so I think we need to keep extending from `ConfigurableService` to keep BC. But, although it extends from deprecated classes, we can try to avoid coupling it even more with the legacy `ConfigurableService` & `ServiceManager` classes while keeping backward compatibility with the former way of passing `$options`.

I'm using the singleton to get the logger by default, but I'm passing a mock for it in tests. Making the constructor to receive a `LoggerInterface` (instead of getting it from the `LoggerService` or the DI container) also serves to try to avoid coupling this class even more with the legacy `ConfigurableService` it extends from.

However, we may pass the logger in the `$options` array as well or get it from the service manager (and tweak a bit the unit tests to mock the SM deps) to lower the changes done in this class. Do you think that would be better?

hectoras: _(In fact, the sole purpose of the changes in this class is to have the logger passed though the constructor, so technically they can be reverted without affecting the fix)_

model/qti/metadata/ontology/GenericLomOntologyClassificationInjector.php
hectoras: It seems `AbstractMetadataService` gets the injectors though the service locator, and right now it does not inject an usable logger for this class (if I make `$logger` non-nullable, the call to the constructor fails, and I've found no easy way to fix it).

model/qti/metadata/ontology/OntologyMetadataInjector.php
hectoras: No AFAIK :disappointed:  see https://github.com/oat-sa/extension-tao-itemqti/pull/2155/files#r899996806

model/qti/metadata/ontology/OntologyMetadataInjector.php
```diff
@@ -26,53 +25,67 @@
...
+
     public function inject($target, array $values)
     {
-        if (!$target instanceof core_kernel_classes_Resource) {
-            $msg = "The given target is not an instance of core_kernel_classes_Resource.";
-            throw new MetadataInjectionException($msg);
-        }
+        $this->assertIsResource($target);
+
+        $this->debug('values = %s', var_export($values, true));
```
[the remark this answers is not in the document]
hectoras: My idea was to keep them to ease debugging in the future in case it is needed. However, we may remove some of them (for example, this two in the top), or maybe keep just `'Remove property "%s" resource="%s" language="%s"'` & `'Set property "%s" value="%s" language="%s"'` ?

model/qti/metadata/AbstractMetadataService.php
gabrielfs7: I did not get, this class uses `extends ConfigurableService`, so we can call the container as I mentioned and not need for constructor override.

model/qti/metadata/ontology/OntologyMetadataInjector.php
gabrielfs7: It is not a problem to me, but I would them keep all in the constructor for consistency (and remove static calls if possible)

model/qti/metadata/ontology/OntologyMetadataInjector.php
```diff
@@ -99,57 +113,116 @@ public function inject($target, array $values)
...
             return $rules[$pathKey][0];
         }
-        
+
         return false;
     }
+
+    private function debug(string $message, ...$replacements): void
+    {
+        $this->logger->info(__CLASS__ . ': ' . vsprintf($message, $replacements));
```
gabrielfs7, line 203: We have to take care about adding logs, cloudwatch is on of the services we pay more today in AWS. Lets try to not make it to verbose, output only in case of error or inconsistencies that could help us solve a problem. 

Check the call for this method, please :)

hectoras: I'll remove them once I finish the tests

**TODO**
- [ ] Remove logger calls in `OntologyMetadataInjector`

model/qti/metadata/ontology/OntologyMetadataInjector.php
hectoras: Changed to instantiate it in the class constructor: 8bc8933d9240154247efe800f7b1656e6d711b61

model/qti/metadata/ontology/OntologyMetadataInjector.php
hectoras: Changes in this file have been reverted, so marking this thread as solved.

model/qti/metadata/ontology/OntologyMetadataInjector.php
hectoras: Changes in this file have been reverted, so marking this thread as solved.

model/qti/metadata/ontology/OntologyMetadataInjector.php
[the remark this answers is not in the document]
hectoras: Changes in this file have been reverted, so marking this thread as solved.

model/qti/metadata/ontology/GenericLomOntologyClassificationInjector.php
hectoras: Logger usage and injection has been removed from `GenericLomOntologyClassificationInjector`, so marking this thread as solved.

model/qti/metadata/ontology/GenericLomOntologyClassificationInjector.php
```diff
@@ -55,21 +51,69 @@ public function inject($target, array $values)
...
+                    $previousValues = $target->getPropertyValuesByLg(
+                        $property,
+                        $langCode
+                    );
+
+                    if (($previousValues->count() > 0) && $property->isMultiple()) {
+                        $target->setPropertyValueByLg($property, $value, $langCode);
+                    } else {
+                        $target->editPropertyValueByLg($property, $value, $langCode);
+                    }
```
gabrielfs7, line 72: comment needed, maybe a new method required? 

```
private function savePropertyValue($property, $value, $langCode): void
{
    $previousValues = $target->getPropertyValuesByLg(
        $property,
        $langCode
    );

    if ($previousValues->count() > 0 && $property->isMultiple()) {
        $target->setPropertyValueByLg($property, $value, $langCode);
        
        return;
    }
    
    $target->editPropertyValueByLg($property, $value, $langCode);
}
```

test/unit/model/qti/metadata/ontology/GenericLomOntologyClassificationInjectorTest.php
```diff
@@ -0,0 +1,531 @@
...
+use core_kernel_classes_ContainerCollection;
+use core_kernel_classes_Property;
+use core_kernel_classes_Resource;
+use oat\generis\model\data\Ontology;
+use oat\taoQtiItem\model\qti\metadata\MetadataInjectionException;
+use oat\taoQtiItem\model\qti\metadata\ontology\GenericLomOntologyClassificationInjector;
+use oat\taoQtiItem\model\qti\metadata\simple\SimpleMetadataValue;
+use PHPUnit\Framework\MockObject\MockObject;
+
+class GenericLomOntologyClassificationInjectorTest extends TestCase
```
gabrielfs7, line 37: Thanks for the test! 👏  🚀

gabrielfs7: For this case, we can extend TestCase from PhpUnit, we do not need the one from generis (which is deprecated).

gabrielfs7 (approved):
Good job, thanks! 

test/unit/model/qti/metadata/ontology/GenericLomOntologyClassificationInjectorTest.php
hectoras: You are right, I missed that one. Changed in bbb92640837eca955413da978903b6307c29cce2, thanks! :)

AntonNeviarouski:
QA verified

hectoras:
Released as [v29.5.5](https://github.com/oat-sa/extension-tao-itemqti/releases/tag/v29.5.5).

hectoras:
Backported as [v29.3.2](https://github.com/oat-sa/extension-tao-itemqti/releases/tag/v29.3.2).

PR #2167, opened 2022-07-06
fix: update dependency to tao-core

krampstudio:
Ensure the dependency to tao-core contains the bundler fix https://github.com/oat-sa/tao-core/releases/tag/v50.18.3


Silvia-loza (approved):
LGTM

PR #2170, opened 2022-07-08
feat/AUT-2332 Add language to editor

meefox:
Related to https://oat-sa.atlassian.net/browse/AUT-2332

Main description https://github.com/oat-sa/tao-core/pull/3575

**Description:**
Add languages to inline CKE Editor

**Changes:**
* Added language plugin for CKE editor, disable if feature flag is not provided (currently in debugging state to show always).

Part 2/5


views/js/qtiCreator/editor/ckEditor/htmlEditor.js
```diff
@@ -90,6 +94,10 @@ define([
             });
         }
 
+        if(languagePluginEnabled) {
+            removePlugins.push('language');
```
jsconan, line 98: **thought:** I wonder if it would not be easier/better to do the opposite, say add the plugin when needed instead of removing when not needed.

views/js/qtiCreator/helper/languages.js
```diff
@@ -0,0 +1,95 @@
...
+        if(!legacyLanguagesData) {
+            legacyLanguagesData = languages.reduce((memo ,lang) => {
+                memo[lang.code] = lang.label;
+
+                return memo;
+            }, {});
+        }
+
+        return legacyLanguagesData;
+    }
```
jsconan, line 56: **quibble:** I'm not sure the approach is correct with the caching. A list is given as a parameter, then it is transformed and the result is cached into a private variable. The next time you call the help with another list, it will return the result from the first call...

I'd rather suggest not caching the transformation because performance benefits are limited. Otherwise, the helper would be better off not accepting a parameter, it could act as an alternative to the loader, internally calling the main helper `getList()`.

BTW, the code contains several code-style issues. You'd need to pass it through Prettier.

views/js/qtiCreator/helper/languages.js
```diff
@@ -0,0 +1,95 @@
...
+     */
+     const useCKEFormatting = (languages) => {
+        if(!CKELanguagesData) {
+            CKELanguagesData = languages.map((lang) => {
+                return `${lang.code}:${lang.label}:${lang.orientation}`
+            });
+        }
+
+        return CKELanguagesData;
+    }
```
jsconan, line 74: **quibble:** same remark here regarding the cache.

jsconan (commented):
The overall approach is good.

I have a few remarks though, especially regarding the way to handle the result of the language request.

views/js/qtiCreator/editor/ckEditor/htmlEditor.js
meefox: disabling plugin is a bit easier if checking the file we have already a flow disabling plugins, and it may be better for future when we will have only solar (enabled by default)

bziondik (approved):
- [ ]  New code is covered by tests (if applicable)
- [ ]  Tests are running successfully (old and new ones) on my local machine (if applicable)
- [x]  New code is respecting code style rules
- [x]  New code is respecting best practices
- [x]  New code is not subject to concurrency issues (if applicable)
- [x]  Feature is working correctly on my local machine (if applicable)
- [x]  Acceptance criteria are respected
- [x]  Pull request title and description are meaningful

shaveko (approved):

bziondik (approved):

PR #2182, opened 2022-07-22
Fix/tr 4451/inlinechoice empty label text

olga-kulish:
Related to https://oat-sa.atlassian.net/browse/TR-4451

- [x] apply https://github.com/oat-sa/tao-item-runner-qti-fe/pull/305


oatymart (approved):
Pre-approving, on condition that branch ref will be updated.
Tested in context of https://github.com/oat-sa/nfer-enot-template/pull/67 and https://github.com/oat-sa/tao-item-runner-qti-fe/pull/305#pullrequestreview-1053835389

PR #2186, opened 2022-07-26
fix/AUT-1134/character-restrictions 

meefox:
Related to https://oat-sa.atlassian.net/browse/AUT-1134
Seattle PR for https://github.com/oat-sa/tao-item-runner-qti-fe/pull/307

- [ ] Update NPM to released version of @oat-sa/tao-item-runner-qti


Silvia-loza (commented):
Fix works well and it looks ok in` item runner` however, I've noticed that when setting the response, even if it's set to integer or float, it allows the user (teacher) to enter any string... 
Should this be like that? WDYT?
image

meefox:
Closed due to AC needs to be improved

PR #2188, opened 2022-07-28
fix: do not keep empty tags to prevent FE img parsing issue

pnal:


model/qti/interaction/ImsPortableCustomInteraction.php
```diff
@@ -133,7 +133,7 @@ public function setLibraries($libraries)
 
     public function toArray($filterVariableContent = false, &$filtered = [])
```
andreluizmachado, line 134:
```suggestion
    public function toArray($filterVariableContent = false, &$filtered = []): array
```

model/qti/interaction/ImsPortableCustomInteraction.php
```diff
@@ -150,7 +150,7 @@ public static function getTemplateQti()
     {
         return static::getTemplatePath() . 'interactions/qti.imspci.tpl.php';
     }
-    
+
     protected function getTemplateQtiVariables()
```
andreluizmachado, line 154:
```suggestion
    protected function getTemplateQtiVariables():array
```

bziondik (changes requested):
Doesn't work save

PR #2143, opened 2022-06-01
feat/AUT-76/stylesheet-manager-for-passages

marpesia:
**Related to:** [AUT-76](https://oat-sa.atlassian.net/browse/AUT-76)

**Description:**
- Add a Stylesheet manager for passages and isolate the scope of the CSS rules to the passage.
- The stylesheets attached to a passage should not affect the style of the items the passage is included into

**Changes:**

- Refactor handle functions between taoMediaManager and taoItemQti

:exclamation: Pending Solar Preview, depends on Solar Delivery part :exclamation:

**PR related:**
- https://github.com/oat-sa/extension-tao-mediamanager/pull/404

**How to check:**

- Check Items and Test with passages with stylesheets.
- Check adding passage with custom styles and without it and CSS file. (Without custom styles may have not mainClass on the passage)
- Check previews Terre.

[Kitchen environment on the ticket.](https://oat-sa.atlassian.net/browse/AUT-76?focusedCommentId=190406)

PR #2211, opened 2022-08-30
Fix/TR-4291/Missing class import

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-4192

~~Bring back a removed class import that is still needed.~~
Use the right class for getting the service ID.

Because of this wrong class used, the export and the publication of tests containing styled shared stimulus are failing with the error `Class 'oat\taoQtiItem\model\Export\Stylesheet\FileManagement' not found`.

This was introduced by #2143 

**How to test:**
- create a shared stimulus and add a stylesheet
- create an item and include the shared stimulus
- create a test containing this item
- export the test
- publish the test

**Requirements:**
- You need to install the extension taoMediaManager.
- In order to make possible the inclusion of the stylesheet in a shared stimulus, you will need to add this to `tao/config/tao/client_lib_config_registry.conf.php`:
```
        'services/features' => array(
            'visibility' => array(
                'taoMediaManager/creator/StyleSheetManager' => 'show'
            )
        ),
```

### DIFF
--- a/model/Export/Stylesheet/AssetStylesheetLoader.php
+++ b/model/Export/Stylesheet/AssetStylesheetLoader.php
@@ -26,7 +26,6 @@ use League\Flysystem\FileNotFoundException;
 use League\Flysystem\FilesystemInterface;
 use oat\generis\model\OntologyAwareTrait;
 use oat\oatbox\filesystem\FileSystemService;
-use oat\taoMediaManager\model\fileManagement\FileManagement;
 use oat\oatbox\service\ConfigurableService;
 use oat\taoMediaManager\model\fileManagement\FlySystemManagement;
 use oat\taoQtiItem\model\Export\AbstractQTIItemExporter;
@@ -100,6 +99,6 @@ class AssetStylesheetLoader extends ConfigurableService
         // FIXME this violates the order of dependencies.
         //  taoQtiItem cannot depend on taoMediaManager as it causes a circular dependency
         //  caused by [#1766](https://github.com/oat-sa/extension-tao-itemqti/pull/1766)
-        return $this->getServiceLocator()->get(FileManagement::SERVICE_ID);
+        return $this->getServiceLocator()->get(FlySystemManagement::SERVICE_ID);
     }
 }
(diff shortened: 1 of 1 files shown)


hectoras (approved):

pnal (approved):

model/Export/Stylesheet/AssetStylesheetLoader.php
```diff
@@ -26,6 +26,7 @@
 use League\Flysystem\FilesystemInterface;
 use oat\generis\model\OntologyAwareTrait;
 use oat\oatbox\filesystem\FileSystemService;
+use oat\taoMediaManager\model\fileManagement\FileManagement;
```
gabrielfs7, line 29: I think this removal was in purpose to "enforce" the usage of `FlySystemManagement`. Check the return type of this method `FlySystemManagement` https://github.com/oat-sa/extension-tao-itemqti/pull/2211/files#diff-2e1fa53dc0a5951160dd0268a6376f80f4b93269b73dfc7dcb2a7e13a37f1defR98

Maybe we should instead use `FlySystemManagement::SERVICE_ID` there or change the return type? Cause the previous change seems to have broken the open/closed principle.

jsconan: Thx Gabriel. I should have taken more time to look at the replacement class signature.

bziondik (approved):
From FE side it LGTM :)

jsconan:
**Note**
This pull request only fixes a symptom. The root cause is a wrong implementation that makes this extension depend on taoMediaManager.

This is bad as it breaks the dependency order and introduces a cyclic dependency.
Moreover, it also causes a failure in the unit tests:

```
......................................PHP Fatal error:  Declaration of oat\taoMediaManager\model\fileManagement\FlySystemManagement::getFileStream($link) must be compatible with oat\taoMediaManager\model\fileManagement\FileManagement::getFileStream(string $link): Psr\Http\Message\StreamInterface in /Users/oat/Projects/solar/nextgen-stack/tao/taoMediaManager/model/fileManagement/FlySystemManagement.php on line 70

Fatal error: Declaration of oat\taoMediaManager\model\fileManagement\FlySystemManagement::getFileStream($link) must be compatible with oat\taoMediaManager\model\fileManagement\FileManagement::getFileStream(string $link): Psr\Http\Message\StreamInterface in /Users/oat/Projects/solar/nextgen-stack/tao/taoMediaManager/model/fileManagement/FlySystemManagement.php on line 70
```

PR #2215, opened 2022-09-12
Feature/TR-4575/Allow external dependencies for PCI

jsconan:
Related to: https://oat-sa.atlassian.net/browse/TR-4575

Modify the PCI bundler, allowing for the inclusion of external dependencies via path mapping.

How to test:
- add a `path.json` file into a PCI for including an external dependency like an npm package
- require the external dependency from the PCI
- bundle the PCI (`$(cd tao/views/build && npx grunt portableelement -e=taoExtension -i=myPci)` )
- check the bundle task performs without error and that the external dependency is present inside the bundle

Example:
- `path.json`:
```json
{
  "external"   : "../../../taoExtension/views/node_modules/something/dist/external"
}
```

- `taoExtension/views/node_modules/something/dist/external.js`:
```javascript
define(function() {
    return () => alert('It works!');
});
```

- PCI:
```javascript
require([..., 'external'], function(..., external) {
    ...
    external();
    ...
});
```

### DIFF
--- a/views/build/grunt/portableelement.js
+++ b/views/build/grunt/portableelement.js
@@ -275,6 +275,11 @@ module.exports = function (grunt) {
                         config.paths[extension] = `${root}/${extension}/views/js`;
                         config.paths[model.id] = model.basePath;
 
+                        // Add path to external dependencies
+                        const pathFile = join(model.basePath, 'paths.json');
+                        if (grunt.file.exists(pathFile)) {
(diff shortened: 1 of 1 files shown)


jsconan:
Unfortunately, this does not work well with the [PCI development mode](https://github.com/oat-sa/taohub-articles/blob/master/forge/pci-development.md).

Actually, we can include "external" module without the need to rely on a mapping file. Instead of using an alias, we can use a deep URI like `myPCI/node_modules/something/dist/external`

Closing this for now.

PR #2220, opened 2022-09-14
Fix/TR-4575/Parsing error with self closing tags in a PCI markup

jsconan:
### Summary

When trying to save an existing item with a PCI having line breaks in the markup, the server returns with the error `Item XML is not valid`.

This is due to an option allowing the XML parser to auto close self closing tags.

### Details

The option `LIBXML_NOEMPTYTAG` was activated for PCI and PIC. As per the [PHP documentation](https://www.php.net/manual/en/libxml.constants.php), this option expands empty tags. As an example `<br />` becomes `<br></br>` and the browser transforms it to `<br /></br>`. When this is sent to the server, the XML validator returns with the error message `Item XML is not valid: Opening and ending tag mismatch`.

The activation of this option was introduced by the PR #317. Unfortunately, this PR does not explain the intent, and I did not find any reference to this in our records.

### How to test

- have an item with a PCI
- edit the prompt of the PCI, adding line breaks
- save the item and exit the authoring
- edit again the item
- save it...

### DIFF
--- a/model/qti/interaction/PortableCustomInteraction.php
+++ b/model/qti/interaction/PortableCustomInteraction.php
@@ -52,27 +52,27 @@ class PortableCustomInteraction extends CustomInteraction
     protected $typeIdentifier = '';
     protected $entryPoint = '';
     protected $version = '0.0.0';
-    
+
     public function setTypeIdentifier($typeIdentifier)
     {
         $this->typeIdentifier = $typeIdentifier;
     }
-    
+
     public function setEntryPoint($entryPoint)
     {
         $this->entryPoint = $entryPoint;
     }
-    
+
     public function getTypeIdentifier()
     {
         return $this->typeIdentifier;
     }
-    
+
     public function getEntryPoint()
     {
         return $this->entryPoint;
     }
-    
+
     public function getProperties()
     {
         return $this->properties;
@@ -91,22 +91,22 @@ class PortableCustomInteraction extends CustomInteraction
     {
         return $this->stylesheets;
     }
-    
+
     public function setStylesheets($stylesheets)
     {
         $this->stylesheets = $stylesheets;
     }
-    
+
     public function getMediaFiles()
     {
         return $this->mediaFiles;
     }
-    
+
     public function setMediaFiles($mediaFiles)
     {
         $this->mediaFiles = $mediaFiles;
     }
-    
+
     public function getVersion()
     {
         return $this->version;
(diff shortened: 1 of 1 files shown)


model/qti/interaction/PortableCustomInteraction.php
```diff
@@ -232,7 +232,7 @@ public function feed(ParserFactory $parser, DOMElement $data, QtiNamespace $xmln
 
         $markupNodes = $parser->queryXPathChildren(['portableCustomInteraction', 'markup'], $data, $xmlnsName);
         if ($markupNodes->length) {
-            $markup = $parser->getBodyData($markupNodes->item(0), true, true);
+            $markup = $parser->getBodyData($markupNodes->item(0), true);
```
gabrielfs7, line 235: I saw this true was added by Bertrand 7 years ago to fix pic/pic. Do you know if this wont bring this problem back?

https://github.com/oat-sa/extension-tao-itemqti/commit/7811fa6abd547c201b6267dcb1963fdcf4168555#diff-8f29b0ca7697b7fe1fcf6b3518e7f8c86076ef8b49b9afb3c12ff98e285ce254R159

jsconan: This is what I explained in the PR description. I did not find any reason for this. And for me, it creates more problems than it can solve.

The problem is that the XML validation does not support it after the browser has passed over the content. The browser accepts short notation for standalone tags, and automatically replaces them with the complete format (`<br>` is interpreted as `<br />`, then `</br>` does not exist, so it ignores it). This produces a broken XML in return. I could have added a filter to eliminate the dead tags before sending the data to the server, but it would add unnecessary complexity. I think it is better to remove a too greedy fix than add another layer of complexity which could introduce more issues. If the problem that was fixed 7 years ago arises again, I'm pretty sure we can fix it in a better way. But for now, I don't see what it could be, honestly.
Anyway, let me do some additional checks with various browsers. Maybe one is still not reformating the standalone tags to auto-closed ones, and in this case, it may also be an issue. But then we would need to apply a proper fix that is not toggling in one way or the other.

gabrielfs7: Ok, lets go with this way and ask QA to validate the pic/pic case :)

gabrielfs7 (approved):
Thanks JS, 

Approved, but lets also ask QA to validate the pic/pic case :)

jsconan:
As I'm not sure the proposed solution is safe enough, I'm closing the pull request.

PR #2246, opened 2022-10-11
fix: Don't extract QTI info for ItemUpdatedEvents triggered by RDF Items with no QTI Data

hectoras:
**Associated Jira issue:** [AUT-2688](https://oat-sa.atlassian.net/browse/AUT-2688)


model/qti/event/UpdatedItemEventDispatcher.php
```diff
@@ -38,14 +38,24 @@ class UpdatedItemEventDispatcher extends ConfigurableService
     private const OBJECT_ELEMENT_REFERENCES_KEY = 'objectElementReferences';
     private const IMG_ELEMENT_REFERENCES_KEY = 'imgElementReferences';
 
-    public function dispatch(Item $qtiItem, core_kernel_classes_Resource $rdfItem): void
+    public function dispatch(?Item $qtiItem, core_kernel_classes_Resource $rdfItem): void
```
gabrielfs7, line 41: Some points: 

- It is not recommended to have the an initial parameter mandatory.
- Does not make sense to call this method if there is not Item as parameter.
- Why are we calling this without item, should not we change the place where this is happening to not call it anymore or generate the item object properly?

It seems we are not fixing the root cause of this problem, but a side effect.

gabrielfs7 (changes requested):
We should fix by providing updated Item as parameter instead

model/qti/event/UpdatedItemEventDispatcher.php
hectoras: This change is for supporting cases when we **don't** have a QTI item; that is, to restore the state back to a point in which the item was created in the interface (so there _is_ a RDF item) but no authoring activity has happened yet (so _no QTI data_ is present).

As per my understanding, triggering the event under such circumstances is needed in order to notify about the change. In this case, to trigger an update event with empty reference arrays in order for listeners to know about the last state (no content = no references).

Having an RDF item with `null` Qti information seems to be valid, and the method from taoQtiItem that provides it to our caller controls such situation by explicitly returning `null`.

(I don't like having the first argument as nullable, but I was trying to keep BC: Allowing `null` as the first parameter won't break existing clients expecting it to be non-nullable).

gabrielfs7: We still should not do it as a workaround otherwise we are now with a really weird method signature that will confuse clients of this class.

I see the place that it is happening that it will be null is https://github.com/oat-sa/extension-tao-revision/blob/master/model/RepositoryService.php#L242 

So, we might fix this place only instead. The other 2 places (QTICreator and ItemImporter) are validating the QTI XML before load (throwing and exception), so it should never reach as null to this method.

hectoras: I think we still need to dispatch the event if no QTI Item is present, don't we? I mean, even if we don't have a Qti item in the restored state, we need to trigger an event associated with the RDF resource URI but with no references set (so listeners can remove the info they had about the previous references).

If that's the case, I think we may:
- Add an additional method to this dispatcher that just receives the RDF item (`dispatchResourceUpdated($rdfItem)`?), and
- change the caller in taoRevision to use that one if it gets a `null` QTI Item.

Would that be better?

gabrielfs7: Is it possible to create a dummy Item in the Revision extension with empty attributes?

hectoras: yes, it is. I'll rollback these changes (and keep only the tests), and open a PR in taoRevision.

hectoras: https://github.com/oat-sa/extension-tao-revision/pull/139

hectoras:
Merging in order to introduce the additional unit tests.

```
root@19f5cc866d02:/var/www/html/taoQtiItem# ../vendor/bin/phpunit -c test/unit/phpunit.xml \
                                                                  --bootstrap ./test/unit/bootstrap.php \
                                                                  test/unit/model/qti/event/
PHPUnit 8.5.28 #StandWithUkraine

.                                                                   1 / 1 (100%)

Time: 50 ms, Memory: 4.00 MB

OK (1 test, 2 assertions)
root@19f5cc866d02:/var/www/html/taoQtiItem# 
```

PR #2262, opened 2022-11-09
fix: catch not found exception

pnal:
When importing item with not yet known PCI the import should not break cause it is a valid case. The problem was an exception which been not intercepting.

### DIFF
--- a/model/portableElement/storage/PortableElementRegistry.php
+++ b/model/portableElement/storage/PortableElementRegistry.php
@@ -508,8 +508,11 @@ abstract class PortableElementRegistry implements ServiceLocatorAwareInterface
      */
     protected function getZipLocation(PortableElementObject $object)
     {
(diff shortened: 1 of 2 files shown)


model/portableElement/storage/PortableElementRegistry.php
```diff
@@ -306,7 +306,11 @@ public function getLatestVersion($identifier)
 
     public function getLatestCompatibleVersion(string $identifier, string $targetVersion): ?PortableElementObject
     {
-        $registered = $this->getAllVersions($identifier);
+        try {
+            $registered = $this->getAllVersions($identifier);
+        } catch (PortableElementNotFoundException $e) {
+            return null;
```
andreluizmachado, line 312: Can we log it at least?

andreluizmachado: Can we have unit tests?

bartlomiejmarszal (changes requested):
Should we log error when catching this exception?

bartlomiejmarszal (approved):

andreluizmachado (approved):

PR #2289, opened 2022-12-02
fix: overwrite translations files

shaveko:
related to https://oat-sa.atlassian.net/browse/INF-155

### Description 

Overwriting the .po file by the translation strings provided by customer


bziondik (approved):
LGTM

PR #2313, opened 2023-01-06
fix: allow to inject new namespace for PCI

siwane:
**Ticket**
https://oat-sa.atlassian.net/browse/OATSD-2274
https://oat-sa.atlassian.net/browse/AUT-2683
https://oat-sa.atlassian.net/browse/LSI-1594

**Description**
While moving PCI from IMS to OAT, migration script [try to inject new namespace](https://github.com/oat-sa/extension-pcisample/blob/master/model/LegacyPciHelper/Task/UpgradeTextReaderInteractionTask.php#L165) to PciObject, unfortunately objects are loaded from namespace. For old PCI, it results on PortableCustomInteraction object and not ImsPortableCustomInteraction.
This PortableCustomInteraction has template with [hardcoded namespace](https://github.com/oat-sa/extension-tao-itemqti/compare/fix/OATSD-2274/remove-hardcoded-pci-namespace?expand=1#diff-fec7f9e8b135c401b3e96e879ffab5db52208c7c75514c5752a46ea2c593b245L23) value that does not allow to update it 😓 

**How to test**
Create item with TextReader and image in it, set PCI namespace within item files.
```xml
<portableCustomInteraction xmlns="http://www.imsglobal.org/xsd/portableCustomInteraction">
```
Launch the command:
```shell
php index.php `\oat\pciSamples\model\LegacyPciHelper\LegacyTextReaderItemUpdate'
```
Item should be updated with _v1 namespace suffix
```xml
<portableCustomInteraction xmlns="http://www.imsglobal.org/xsd/portableCustomInteraction_v1">
```

### DIFF
--- a/model/qti/templates/interactions/qti.portableCustomInteraction.tpl.php
+++ b/model/qti/templates/interactions/qti.portableCustomInteraction.tpl.php
@@ -1,5 +1,5 @@
 <?php
-/*
+/**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * as published by the Free Software Foundation; under version 2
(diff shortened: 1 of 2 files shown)


mianchenia (approved):
LGTM. 

Could we avoid so much code-style changes for hotfix PR?

siwane:
I understand your concern but CI is breaking if PSR-12 is not met for all files touched

model/qti/templates/interactions/qti.portableCustomInteraction.tpl.php
```diff
@@ -1,46 +1,61 @@
...
-                <?php foreach(get_data('stylesheets') as $stylesheet):?>
-                <link href="<?=$stylesheet?>" type="text/css" title="base"/>
-                <?php endforeach;?>
-            </stylesheets><?php endif;?>
-            <?php if(count(get_data('mediaFiles'))):?><mediaFiles>
-                <?php foreach(get_data('mediaFiles') as $file):?>
-                <file src="<?=$file?>"/>
-                <?php endforeach;?>
-            </mediaFiles><?php endif;?>
+            <?php if (count(get_data('libraries'))) :?>
```
pnal, line 34: This refactoring may change line formatting in outcome xml making it less readable. 
If you changed it to avoid code format warnings in CI only, I would suggest adding `// phpcs:ignoreFile` comment

alroniks (approved):
I've tested it locally and it works as described. LGTM.

PR #2339, opened 2023-04-10
feature: check for interaction choice validity in pair association widget

albertosgz:
Ticket: https://oat-sa.atlassian.net/browse/AUT-2716

- What's Changed

Check for choice object before get its ID to avoid null access.

- How to test (took from ticket)

1. Go to Items tab
2. Choose an item from Preconditions and click ‘Preview’
3. Drag and drop 1 choice containing the media and one without it into the empty association pair
Actual result: Pair is created but user cannot create more pairs, empty pair doesn’t appear. Several errors appear in browser console, the last dragged choice is stuck with mouse cursor.
Expected result: Association pair is added, empty pair appears so it is possible to create a new association pair. No errors in DevTools console.


views/package.json
```diff
@@ -9,6 +9,6 @@
     },
     "dependencies": {
         "@oat-sa/tao-item-runner": "0.8.1",
-        "@oat-sa/tao-item-runner-qti": "1.9.2"
+        "@oat-sa/tao-item-runner-qti": "dev-fix/AUT-2716/be-able-to-associate-pairs-with-audio-video-in-association-interaction as 1.9.2"
```
bziondik, line 12:
```suggestion
        "@oat-sa/tao-item-runner-qti": "https://github.com/oat-sa/tao-item-runner-qti-fe#fix/AUT-2716/be-able-to-associate-pairs-with-audio-video-in-association-interaction"
```
https://docs.npmjs.com/cli/v8/configuring-npm/package-json#git-urls-as-dependencies

PR #2354, opened 2023-05-03
chore: apply PSR-12

shpran:
# [ADF-1440](https://oat-sa.atlassian.net/browse/ADF-1440)

[ADF-1440]: https://oat-sa.atlassian.net/browse/ADF-1440?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


model/qti/choice/Choice.php
```diff
@@ -88,8 +90,9 @@ public function isIdentifierAvailable($newIdentifier)
                 $collection = $relatedItem->getIdentifiedElements();
 
                 try {
-                    $uniqueChoice = $collection->getUnique($newIdentifier, 'oat\\taoQtiItem\\model\\qti\\choice\\Choice');
-                    $uniqueOutcome = $collection->getUnique($newIdentifier, 'oat\\taoQtiItem\\model\\qti\\OutcomeDeclaration');
+                    $uniqueChoice = $collection->getUnique($newIdentifier, Choice::class);
```
gabrielfs7, line 93: Are you sure that `Choice::class` === `oat\\taoQtiItem\\model\\qti\\choice\\Choice` ? 
I would avoid any type of change like this to reduce chances for bugs.

model/qti/choice/Choice.php
```diff
@@ -88,8 +90,9 @@ public function isIdentifierAvailable($newIdentifier)
...
+                    $uniqueOutcome = $collection->getUnique($newIdentifier, OutcomeDeclaration::class);
```
gabrielfs7, line 94: Same here https://github.com/oat-sa/extension-tao-itemqti/pull/2354/files?diff=split&w=1#r1190676860

model/qti/interaction/MatchInteraction.php
```diff
@@ -161,7 +162,7 @@ public function createChoice($choiceAttributes = [], $choiceValue = null, $setNu
         $returnValue = null;
 
         if ($this->isValidMatchSetNumber($setNumber)) {
-            if (!empty(static::$choiceClass) && is_subclass_of(static::$choiceClass, 'oat\\taoQtiItem\\model\\qti\\choice\\Choice')) {
+            if (!empty(static::$choiceClass) && is_subclass_of(static::$choiceClass, Choice::class)) {
```
gabrielfs7, line 165: Same as here https://github.com/oat-sa/extension-tao-itemqti/pull/2354/files?diff=split&w=1#r1190676860

helpers/CssHelper.php
```diff
@@ -24,11 +24,12 @@
 
 class CssHelper
 {
-
+    // phpcs:disable PSR2.Methods.MethodDeclaration
     private static function _buildWarning(): string
     {
         return " /* Do not edit */";
     }
+    // phpcs:enable PSR2.Methods.MethodDeclaration
```
hectoras, line 32: This is a private method, can we rename the method as `buildWarning()` and fix the call on line 47 instead of disabling the warning? :)

model/qti/IdentifierCollection.php
```diff
@@ -75,44 +75,46 @@ public function add(IdentifiedElement $element)
             $this->elements[$identifier][$element->getSerial()] = $element;
         }
     }
-    
+
     public function exists($identifier)
     {
-        
+
         $returnValue = false;
```
hectoras, line 82:
```suggestion
        $returnValue = false;
```

model/qti/IdentifierCollection.php
```diff
@@ -75,44 +75,46 @@ public function add(IdentifiedElement $element)
...
+
         return $returnValue;
     }
-    
+
     public function get($identifier = '')
     {
-        
+
         $returnValue = [];
```
hectoras, line 96:
```suggestion
        $returnValue = [];
```

model/qti/IdentifierCollection.php
```diff
@@ -75,44 +75,46 @@ public function add(IdentifiedElement $element)
...
+
         return $returnValue;
     }
-    
+
     public function getUnique($identifier, $elementClass = '')
     {
-        
+
         $returnValue = null;
```
hectoras, line 110:
```suggestion
        $returnValue = null;
```

model/qti/IdentifierCollection.php
```diff
@@ -151,10 +153,10 @@ public function addMultiple($identifiedElements)
             }
         }
     }
-    
+
     public function merge(IdentifierCollection $identifierCollection)
     {
-        
+
         foreach ($identifierCollection->get() as $elements) {
```
hectoras, line 160:
```suggestion
        foreach ($identifierCollection->get() as $elements) {
```

model/qti/InfoControl.php
```diff
@@ -67,12 +66,12 @@ public function setMarkup($markup)
 
     public function toArray($filterVariableContent = false, &$filtered = [])
     {
-        
+
         $returnValue = parent::toArray($filterVariableContent, $filtered);
```
hectoras, line 70:
```suggestion
        $returnValue = parent::toArray($filterVariableContent, $filtered);
```

model/qti/InfoControl.php
```diff
@@ -83,18 +82,18 @@ public static function getTemplateQti()
 
     protected function getTemplateQtiVariables()
     {
-        
+
         $variables = parent::getTemplateQtiVariables();
```
hectoras, line 86:
```suggestion
        $variables = parent::getTemplateQtiVariables();
```

model/qti/InfoControl.php
```diff
@@ -83,18 +82,18 @@ public static function getTemplateQti()
...
+
         return $variables;
     }
-    
+
     public function feed(ParserFactory $parser, DOMElement $data, QtiNamespace $xmlns = null)
     {
-        
+
         $markup = $parser->getBodyData($data->item(0), true);
```
hectoras, line 97:
```suggestion
        $markup = $parser->getBodyData($data->item(0), true);
```

install/update/class.taoQtiUpdate.php
```diff
@@ -73,9 +74,9 @@ protected function convertQtiItem(core_kernel_classes_Resource $item)
 
     protected function convertQtiFromV2p0ToV2p1($xml)
     {
-        
+
         $returnValue = '';
```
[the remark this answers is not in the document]
shpran: In scope of current task, I don't want to fix things not related to PSR-12 🙂 
This will require additional time to find all of them.

But thanks for this 👍🏻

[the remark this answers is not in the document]
hectoras: There is a whiteline between the opening bracket for the method and the method body itself, and I think that line should be removed.

PSR-12 does not explicitly say that "the function body must not start with a whiteline", but IMO that is inferred from the examples in the specification. Also, PSR-2, which PSR-12 extends from, shows warnings for "Blank line found at start of control structure" ([ref](https://youtrack.jetbrains.com/issue/WI-41602)), and [section 5 of PSR-12](https://www.php-fig.org/psr/psr-12/#5-control-structures) states that "The body MUST be on the next line after the opening brace".

BTW I don't consider this a blocking issue for this PR to pass :slightly_smiling_face:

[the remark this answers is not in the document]
gabrielfs7: @hectoras we are out of budget/time for the PSR-12, so if the build passes and the code is not broken, we can approve the PRs :)

gabrielfs7 (approved):

install/update/class.taoQtiUpdate.php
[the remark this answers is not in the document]
hectoras: Yeah, I understand, I've already approved it.

PR #2356, opened 2023-05-08
fix/AUT-3093-math-overflow

rgomeztao:
coming from [https://oat-sa.atlassian.net/browse/AUT-3093](https://oat-sa.atlassian.net/browse/AUT-3093)

changelog:
* update style to remove selector "has"

how to test:
http://test-rgomez1.playground.kitchen.it.taocloud.org:42051/


shaveko (changes requested):
⚠️ your PR targets master branch!

PR #2402, opened 2023-09-19
fix: add compiled translation files

bartlomiejmarszal:
Add missing compiled js file for translations


KirylHatalski (approved):

pribi (approved):
I cannot properly check the diff. What is changed exactly?
Isn't this file generated?

bartlomiejmarszal:
> I cannot properly check the diff. What is changed exactly?
> 
> Isn't this file generated?

Yes. It should be included with a previous commit with .po file

PR #2407, opened 2023-10-03
feat: place imported assets to correct class

pnal:
Ticket: https://oat-sa.atlassian.net/browse/PISA25-561
## What's Changed

- Passages related to imported items will be placed to classes with path `{test name}/{item name}` or `{target class}/{test name}/{item name}` in case if test is imported by API [/taoQtiTest/RestQtiTests/importDeferred](https://github.com/oat-sa/extension-tao-testqti/blob/master/doc/swagger.json#L347) and `itemClassUri` specified.

## How to test
- Import test with passage via UI
- Import test with passage via API
- Check path where passage assets saved

### DIFF
--- a/model/qti/ImportService.php
+++ b/model/qti/ImportService.php
@@ -94,6 +94,8 @@ class ImportService extends ConfigurableService
 
     public const PROPERTY_QTI_ITEM_IDENTIFIER = 'http://www.tao.lu/Ontologies/TAOItem.rdf#QtiItemIdentifier';
 
+    private const MEDIA_ROOT_CLASS = 'http://www.tao.lu/Ontologies/TAOMedia.rdf#Media';
+
     /**
      * @return ImportService
      */
@@ -577,12 +579,7 @@ class ImportService extends ConfigurableService
                         ->get(\common_ext_ExtensionsManager::SERVICE_ID)
                         ->isInstalled('taoMediaManager')
                 ) {
-                    // Media manager path where to store the stimulus.
-                    $path = $this->retrieveFullPathLabels($itemClass);
-                    // Uses the first created item's label as the leaf class.
-                    if (is_array($path)) {
-                        array_unshift($path, $rdfItem->getLabel());
-                    }
+                    $mediaClass = $this->getTargetClassForAssets($itemClass, $rdfItem);
                     /** Shared stimulus handler */
                     $sharedStimulusHandler = new SharedStimulusAssetHandler();
                     $sharedStimulusHandler->setServiceLocator($this->getServiceLocator());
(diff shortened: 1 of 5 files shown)


swiec:
@mike-ionut-mihai-sandu-tao just to remind you, that you still didn't pay my invoice. I hope you will start responding to my emails

model/qti/ImportService.php
```diff
@@ -94,6 +94,8 @@ class ImportService extends ConfigurableService
 
     public const PROPERTY_QTI_ITEM_IDENTIFIER = 'http://www.tao.lu/Ontologies/TAOItem.rdf#QtiItemIdentifier';
 
+    private const MEDIA_ROOT_CLASS = 'http://www.tao.lu/Ontologies/TAOMedia.rdf#Media';
```
bartlomiejmarszal, line 97: This const live here: `tao/taoMediaManager/model/TaoMediaOntology.php:27`
But it is in the MediaManager namespace. This raises a concern that it is not a proper place for this code. Media are not a part of itemQti extension and it should not be handle here

bartlomiejmarszal (changes requested):

model/qti/ImportService.php
```diff
@@ -883,28 +880,33 @@ protected function getMetadataImporter()
...
      * Retrieve the labels of all parent classes up to base item class.
-     *
-     * Return values: an empty string if the $class parameter is not a class object
-     *                an empty array if the $class parameter is the base item class
-     *                an array in other cases, ordered from base item class to the class given as parameter
-     *
-     * @param mixed $class Class from which to retrieve.
-     * @return array|string
      */
-    public function retrieveFullPathLabels($class)
```
bartlomiejmarszal, line 894: Why did you remove this public method? Are you sure it's not used anywhere?

bartlomiejmarszal (changes requested):
**Review Checklist**
- [ ] Pull request's target is not `master`
- [ ] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [ ] Commits messages are meaningful
- [ ] Commits are atomic

model/qti/ImportService.php
pnal: This constant was used in the process anyway I just moved it in another place. The issue of coupling media manager is bigger than just this const, to decouple it correctly we need a huge refactoring. Other option is to give up and consider it as a dependency. To be precise this PR is more a fix than a feature.

model/qti/ImportService.php
```diff
@@ -883,28 +880,33 @@ protected function getMetadataImporter()
...
-    {
-        if (!$class instanceof core_kernel_classes_Class) {
-            return '';
+    public function getTargetClassForAssets(
+        core_kernel_classes_Class $itemClass,
+        core_kernel_classes_Resource $itemResource
+    ): core_kernel_classes_Class {
+        // Collecting labels path from item root to the class where the item resource is stored
+        $labels = [];
+        while ($itemClass->getUri() !== TaoOntology::CLASS_URI_ITEM) {
```
Karol-Stelmaczonek, line 890: I'm worried that we could end here with infinite loop? I don't know if class could reference to itself or something?

Karol-Stelmaczonek (approved):

model/qti/ImportService.php
pnal: it's the very old code which has never cause any problem, covered by tests, if we had RDF tree broken it here would be a last problem we should be worried about.

bartlomiejmarszal (approved):

PR #2410, opened 2023-10-11
remove lodash if possible

taopkorczak:
Related: https://oat-sa.atlassian.net/browse/ADF-1579

Remove lodash usage if possible

### DIFF
--- a/views/js/QtiPreviewResultServerApi.js
+++ b/views/js/QtiPreviewResultServerApi.js
@@ -18,10 +18,9 @@
  */
 define([
     'jquery',
-    'lodash',
     'taoQtiItem/qtiCommonRenderer/helpers/PciResponse',
     'util/strPad'],
-    function($, _, pciResponse, strPad){
+    function($, pciResponse, strPad){
     'use strict';
 
     function QtiPreviewResultServerApi(endpoint, itemUri){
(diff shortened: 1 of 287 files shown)


views/js/QtiPreviewResultServerApi.js
```diff
@@ -44,9 +43,15 @@ define([
         var previewConsole = $('#preview-console');
         var variableIdentifier;
 
+        function escapeHtml(text) {
+            const div = document.createElement("div");
+            div.textContent = text;
+            return div.innerHTML;
+        }
```
oatymart, line 50: I realise you were tasked with removing lodash as much as possible, but I feel that adding local helpers like this is a code smell. (This one is a relatively simple function, so not the worst example.)

If we must have utility functions instead of lodash imports, shouldn't they all live in a common place?

I'm not even sure that the full removal is the smartest step, if it makes us end up with 10-20 "lodash-equivalent" helpers that we have to test and maintain.

I would prefer to see a PR that focuses first on removing just the incompatible usages between lodash 2 and 4 (I don't know how big that change would be, I'm guessing 15% of the size), and then we could go further with the vanilla Javascript replacements if desired.

Overall, this PR is too big to be properly reviewed. The changes could have been done in separate commits (first all the `_.each()`, then all the `_.assign()` etc.)

views/js/legacyPortableSharedLib/OAT/util/EventMgr.js
```diff
@@ -1,18 +1,21 @@
...
-            if(event && events[event]){
-                return _.clone(events[event]);
-            }else{
-                return [];
+            if (event && events[event]) {
+                if (Array.isArray(events[event])) {
+                    return events[event].slice();
+                } else if (typeof events[event] === 'object' && events[event] !== null) {
+                    return Object.assign({}, events[event]);
+                }
```
oatymart, line 14: Same concern for `_.clone()` replacement.

PR #2421, opened 2023-11-22
fix: Wrong shuffle feature id

yaraslau-kavaliou:
Related incident - https://oat-sa.atlassian.net/browse/OATSD-2958

After the fix the `shuffle` feature is enabled for Simple Choice interaction even though `allowElimination` feature is disabled.

Screenshot from 2023-11-22 12-46-09


siwane:
Master PR :p !
Do you know which config it relies on?

siwane (approved):

PR #2472, opened 2024-03-25
Fix null deprecated message for trim 

olegkhuss:
When installing TAO on PHP 8.1 stack We get lot of [passing null deprecation](https://php.watch/versions/8.1/internal-func-non-nullable-null-deprecation) message

This PR aims to fix it to make TAO installation a better


model/qti/OutcomeDeclaration.php
```diff
@@ -69,7 +69,7 @@ protected function getTemplateQtiVariables()
         $variables = parent::getTemplateQtiVariables();
         $variables['defaultValue'] = null;
         $defaultValue = $this->getDefaultValue();
-        if (!is_null($defaultValue) || trim($defaultValue) != '') {
+        if (!is_null($defaultValue) && trim($defaultValue) != '') {
```
vbyndych, line 72: **praise**: good catch!

PR #2493, opened 2024-05-10
Fix/aut 3653/calendar password properties and remote lists

bartlomiejmarszal:
This change will fix:
- calendar properties export / injection
- password properties export / injection
- remote list properties export / injection


Karol-Stelmaczonek (commented):
Please fix the unittests

PR #2501, opened 2024-05-22
Fix/aut 3654/prevent property duplication widget added to mapping

bartlomiejmarszal:
Guard property import with widget type used for input value for property.

Prevent from property value duplication for resource that was causing resource to be displayed multiple times in tree view.

How to test:
1. Create property with i.e password type on one instance
2. Create resource with new property defined
3. Export resource
4. On another instance create class with property with equal type (i.e password)
5. On another instance create class with property with different type (i.e text field)
6. Import resource on different instance to class with equal property type (i.e password) 
**Expected Behaviour:** Resource will be imported
7. Import resource on different instance to class with different property type (i.e text field)
**Expected Behaviour:** Import will fail


tikhanovichA:
Could we also fix unit tests? 
There were 2 errors:

1) oat\taoQtiItem\test\unit\model\qti\metadata\importer\MetaMetadataImportMapperTest::testMapMetaMetadataToProperties Undefined array key "widget"
2) oat\taoQtiItem\test\unit\model\qti\metadata\ontology\MappedMetadataInjectorTest::testInject
TypeError: count(): Argument #1 ($value) must be of type Countable|array, null given

tikhanovichA (approved):

PR #2507, opened 2024-05-30
fix: disable metadata export for APIP

bartlomiejmarszal:
When exporting APIP package, disable metadata export. 

Metadata export for APIP will be developed soon so it will be enabled. 

How to test:
1. Export Package with APIP Content Package
2. Unzip and open file. Make sure metadata has not been included. 
3. Import APIP package

Expected behavior
Item APIP package does not contain metadata. 
Importing APIP Package is successful.

### DIFF
--- a/model/Export/QTIPackedItemExporter.php
+++ b/model/Export/QTIPackedItemExporter.php
@@ -199,11 +199,14 @@ class QTIPackedItemExporter extends AbstractQTIItemExporter
             }
 
             $manifest = $this->getManifest();
-            $this->getMetadataExporter()->export($this->getItem(), $manifest);
-            $this->injectMetadataToManifest($manifest, $this->getItem());
+
+            //APIP does not support metadata at this moment
+            if(!isset($options['apip']) && $options['apip'] !== true){
+                $this->getMetadataExporter()->export($this->getItem(), $manifest);
+                $this->injectMetadataToManifest($manifest, $this->getItem());
+            }
+
             $this->setManifest($manifest);
-
-
             // -- Overwrite manifest in the current ZIP archive.
             $zipArchive->addFromString('imsmanifest.xml', $this->getManifest()->saveXML());
         } else {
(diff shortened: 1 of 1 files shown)


pnal (approved):

tikhanovichA (approved):

model/Export/QTIPackedItemExporter.php
```diff
@@ -199,11 +199,14 @@ public function exportManifest($options = [], $exportResult = [])
             }
 
             $manifest = $this->getManifest();
-            $this->getMetadataExporter()->export($this->getItem(), $manifest);
-            $this->injectMetadataToManifest($manifest, $this->getItem());
-            $this->setManifest($manifest);
 
+            //APIP does not support metadata at this moment
+            if(!isset($options['apip']) && $options['apip'] !== true){
```
Karol-Stelmaczonek, line 204: Shouldn it be OR instead of AND ? If `apip` is not set OR `apip` has value `false` ?  Then do metadata stuff?

Karol-Stelmaczonek (approved):

model/Export/QTIPackedItemExporter.php
julien-sebire: This is incorrect. With an OR, the evaluation of the second member is done only if the first member is false, which means here that $options['apip'] is set. With AND, the second member will be evaluated only if the first member is true, which means that $options['apip'] is not set. Karol is right, it should be OR instead of AND.

model/Export/QTIPackedItemExporter.php
```diff
@@ -199,11 +199,14 @@ public function exportManifest($options = [], $exportResult = [])
             }
 
             $manifest = $this->getManifest();
-            $this->getMetadataExporter()->export($this->getItem(), $manifest);
-            $this->injectMetadataToManifest($manifest, $this->getItem());
-            $this->setManifest($manifest);
 
+            //APIP does not support metadata at this moment
+            if (!isset($options['apip']) || (isset($options['apip']) && $options['apip'] !== true)) {
```
julien-sebire, line 204: With an OR, the evaluation of the second member is done only if the first member is false, which means here if $options['apip'] is set. So isset($options['apip']) is useless here. It can be reduced to:
if (!isset($options['apip']) || $options['apip'] !== true)

PR #2535, opened 2024-06-27
feat: override item identifier validation when ff is enabled

bartlomiejmarszal:
This will override item identifier validation base on FEATURE_FLAG_UNIQUE_NUMERIC_QTI_IDENTIFIER

### DIFF
--- a/model/FeatureFlag/UniqueNumericQtiIdentifierClientConfig.php
+++ b/model/FeatureFlag/UniqueNumericQtiIdentifierClientConfig.php
@@ -27,7 +27,7 @@ use oat\tao\model\featureFlag\FeatureFlagConfigHandlerInterface;
 
 class UniqueNumericQtiIdentifierClientConfig implements FeatureFlagConfigHandlerInterface
 {
-    public const QTI_ID_PATTERN = '/^\\d{9}$/';
+    public const QTI_ID_PATTERN = '/^[^\t\n\r]*$/';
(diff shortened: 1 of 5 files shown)


model/FeatureFlag/UniqueNumericQtiIdentifierClientConfig.php
```diff
@@ -27,14 +27,15 @@
 
 class UniqueNumericQtiIdentifierClientConfig implements FeatureFlagConfigHandlerInterface
 {
+    public const QTI_ID_PATTERN = '/^\\d{9}$/';
```
Karol-Stelmaczonek, line 30: Is that correct pattern with `\\d` ? Should it be `\d` ?

Karol-Stelmaczonek (approved):

model/FeatureFlag/UniqueNumericQtiIdentifierClientConfig.php
tikhanovichA: Should we update it with Normalize string?

bartlomiejmarszal: It should work. And we should keep it as regexp that will accept 9 digits. It is `UniqueNumericQtiIdentifier` validation.

tikhanovichA (approved):

PR #2545, opened 2024-07-10
fix: broken import form

bartlomiejmarszal:
This will fix fatal error after removing feature flag related to metadata export/import


tikhanovichA (approved):
LGTM
Import/export looks ok now

PR #2562, opened 2024-08-16
Fix/AUT-1141/Unwanted confirm exit dialog

jsconan:
### Summary

Prevent from applying the confirm exit dialog if the click was made inside a detached element.

Also added a fix for removing linked CSS from the resource manager.

### Details

The root cause is the same for several cases when the confirm dialog is shown after a click inside the item authoring. For example when adding/removing a feedback.

When a change is made to the item and a click is detected on the page, the change tracker tries to identify if the clicked element belong to the item creator or is outside. The confirm dialog is shown if the click is considered outside. However, if the clicked element has been detached from the DOM, it becomes impossible to know if it was belonging to the item creator before its removal, and therefore the click is considered as outside…

To fix the issue, we need to discard the detached elements from the check.

### How to test

- checkout the branch: `git checkout -t origin/fix/AUT-1141/unwanted-confirm-exit-dialog`
- author an item and add a stylesheet
- save the item
- remove the stylesheet: no confirm dialog should pop up
- enter in answer mode, then add feedbacks (both if and else): no confirm dialog should pop up
- remove the feedbacks: : no confirm dialog should pop up
- click the add stylesheet button, then delete the stylesheet from the resource manager: it should also remove the stylesheet from the item

### DIFF
--- a/views/js/qtiCreator/editor/styleEditor/styleSheetToggler.js
+++ b/views/js/qtiCreator/editor/styleEditor/styleSheetToggler.js
@@ -91,7 +91,7 @@ define([
                     uploadUrl: itemConfig.fileUploadUrl,
                     deleteUrl: itemConfig.fileDeleteUrl,
                     downloadUrl: itemConfig.fileDownloadUrl,
-                    fileExistsUrl : itemConfig.fileExistsUrl,
+                    fileExistsUrl: itemConfig.fileExistsUrl,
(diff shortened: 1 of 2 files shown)


viktar-dzmitryieu-tao (approved):

shaveko (approved):
Works fine locally!

Good job!

PR #2575, opened 2024-09-13
Feat/hkd 6/integration

gabrielfs7:
https://oat-sa.atlassian.net/browse/HKD-6

# Goal 

- Integration branch

# TODO 

- [ ] Update composer dependencies

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -66,9 +66,9 @@
         "oat-sa/oatbox-extension-installer": "~1.1||dev-master",
         "naneau/semver": "~0.0.7",
         "oat-sa/generis": ">=15.36.4",
-        "oat-sa/tao-core": ">=54.22.0",
-        "oat-sa/extension-tao-item": ">=12.0.0",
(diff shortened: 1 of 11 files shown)


model/import/ChecksumGenerator.php
```diff
@@ -37,9 +37,13 @@ public function __construct(ListService $listService)
     public function getRangeChecksum(Property $property): string
     {
         $labels = [];
-        foreach ($this->listService->getListElements($property->getRange()) as $listEntry) {
-            $labels[] = strtolower($listEntry->getLabel());
+
+        if ($property->getRange()) {//@TODO @FIXME Validate the fix with Bart before merge
```
gabrielfs7, line 41: - [ ] Revert this change after proper fix is addressed by authoring team

bartlomiejmarszal: Here is PR that will prevent passing property without range
https://github.com/oat-sa/extension-tao-itemqti/pull/2587

shpran (approved):

PR #2603, opened 2024-10-30
feat: use backend generated item identifier if provided

shaveko:
related to https://oat-sa.atlassian.net/browse/ADF-1801

### Description

Use backend-provided item identifier if supplied. Use frontend-generated one if not.

### How to test

create new item. enter its authoring, check the identifier is the same as with provided by getItemData backend call


controller/QtiCreator.php
```diff
@@ -369,4 +371,31 @@ private function validateXmlInput(string $xml)
...
+
+        if (!$this->getFeatureFlagChecker()->isEnabled('FEATURE_FLAG_UNIQUE_NUMERIC_QTI_IDENTIFIER')) {
+            return $identifier;
+        }
+
+        $uniqueId = $item->getOnePropertyValue($this->getProperty(TaoOntology::PROPERTY_UNIQUE_IDENTIFIER));
+
+        return $uniqueId
+            ? $uniqueId->literal
+            : $identifier;
```
gabrielfs7, line 389: - [ ] @shpran Should not we ask if the Item has the `TaoOntology::PROPERTY_UNIQUE_IDENTIFIER` already first? In the clone, we should set the proper QTI and Metadata for unique identifier, why would we generate it here?
- [ ] Does it make sense to move this logic to `Service::singleton()->getDataItemByRdfItem` and remove all the logic from this controller? I think it is more clean and controller will not deal with business logic

```suggestion
        if ($this->getFeatureFlagChecker()->isEnabled('FEATURE_FLAG_UNIQUE_NUMERIC_QTI_IDENTIFIER')) {
            $uniqueId = $item->getOnePropertyValue($this->getProperty(TaoOntology::PROPERTY_UNIQUE_IDENTIFIER));
        
            if ($uniqueId) {
                return $uniqueId->literal;
            }
        }

        return $this->getIdentifierGenerator()->generate([
            IdentifierGenerator::OPTION_RESOURCE_ID => $item->getUri(),
        ]);
```

migrations/Version202410311205211101_taoQtiItem.php
```diff
@@ -0,0 +1,66 @@
...
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        /** @var EventManager $eventManager */
+        $eventManager = $this->getServiceManager()->get(EventManager::SERVICE_ID);
+
+        $eventManager->attach(
```
gabrielfs7, line 30: - [ ] @shpran Notice that on `ResourceCreated` event we also deal with that and there is already some listener created for this. We need to make sure we do not have discrepant logic among those event / listeners.

model/qti/identifierGenerator/IdentifierGenerator.php
```diff
@@ -24,5 +24,7 @@
 
 interface IdentifierGenerator
 {
-    public function generate(): string;
+    public const OPTION_RESOURCE_ID = 'resourceId';
+
+    public function generate(array $options = []): string;
```
gabrielfs7, line 29: - [ ] @shpran lets please put some annotations to document the options format here and accepted option

model/qti/identifierGenerator/SimpleQtiIdentifierGenerator.php
```diff
@@ -0,0 +1,51 @@
...
+        if (!is_string($resourceId)) {
+            throw new InvalidArgumentException(
+                'QTI Identifier generation failure: resource ID is required and must be a string'
+            );
+        }
+
+        $prefix = LOCAL_NAMESPACE . '#';
+        $resourceId = $options['resourceId'];
+
+        // If we have an item from the current environment
```
gabrielfs7, line 42: - [ ] @shpran It is generally recommend to move to a private explanatory method, rather than comments
- [ ] If I am not mistaken we have already services that generate the URI id, deal with namespaces, etc. Are we sure we need this logic here? Can't we use existing services for that?

```

```

views/js/qtiCreator/helper/itemLoader.js
```diff
@@ -55,8 +55,13 @@ define([
                 // request doesn't handle empty response with 200 code. See: core/request.js:240
                 const itemRdf = request(config.itemDataUrl, { uri: config.uri }).catch(d => d);
 
+                const containsOnlyIdentifier = (data) => {
```
jsconan, line 58: **question:** is Prettier properly configured? lone parameters should not have parenthesis...

jsconan (approved):
Code review for the FE part only.
It looks good, considering that from now on the item identifier will always be generated server-side.

BTW, is Prettier properly configured? I saw a small glitch.

**Review Checklist**
- [x] New code respects code style rules
- [x] New code respects best practices
- [x] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] Commits messages are meaningful
- [x] Commits are atomic

controller/QtiCreator.php
shpran: This part of code was done only to send ID to fronted before XML creation and will be removed, because we do not need synchronization between property and QTI ID anymore.

migrations/Version202410311205211101_taoQtiItem.php
shpran: I'm not triggering ResourceCreated event

model/qti/identifierGenerator/SimpleQtiIdentifierGenerator.php
shpran: This logic was in private method and was used only for QTI creation.
Also, this will be removed, because we will not sync property with QTI ID, so I can use unique ID generation separately without other generators (just to generate property if FF enabled)

migrations/Version202410311205211101_taoQtiItem.php
gabrielfs7: I think it is triggered by clone functionality

controller/QtiCreator.php
gabrielfs7: We still need the synchronization, for the moment

model/qti/identifierGenerator/SimpleQtiIdentifierGenerator.php
gabrielfs7: For URI logic we already have the `generis/uriProvider` and the `core_kernel_uri_UriService`, lets not duplicate sensitive logic here, please

migrations/Version202410311205211101_taoQtiItem.php
shpran: Yes, but as I mentioned, I did not check for ResourceCreated event here 🙂 
I created new event for this - `InstanceCopiedEvent`, so now I have a specific event for specific action

gabrielfs7 (approved):
Approved, but tests and clean up to be performed in integration branch

Thanks

PR #2604, opened 2024-10-30
fix: add rtl on creating translation

gabrielfs7:
# Goal 

Automatically set dir=rtl for rtl locales


jsconan (approved):
All good, it works well! Good job!

image

image

PR #2607, opened 2024-11-06
fix: unique identifiers

shpran:
Fix for unique identifiers + removing unnecessary classes


model/UniqueId/Service/UniqueIdSetter.php
```diff
@@ -0,0 +1,88 @@
...
+        LoggerInterface $logger
+    ) {
+        $this->featureFlagChecker = $featureFlagChecker;
+        $this->resourceMetadataRetriever = $resourceMetadataRetriever;
+        $this->resourceUniqueIdRetriever = $resourceUniqueIdRetriever;
+        $this->qtiIdentifierSetter = $qtiIdentifierSetter;
+        $this->logger = $logger;
+    }
+
+    public function __invoke(core_kernel_classes_Resource $item): core_kernel_classes_Resource
```
gabrielfs7, line 57: - [ ] It is an object, we must update by reference. 

```suggestion
    public function __invoke(core_kernel_classes_Resource $item): void
```

model/UniqueId/Service/UniqueIdSetter.php
```diff
@@ -0,0 +1,88 @@
...
+            $this->logger->info(sprintf('Cannot retrieve original resource URI for item %s', $item->getUri()));
+
+            return $item;
+        }
+
+        $originalResource = $item->getResource($originalResourceUri);
+        $uniqueIdentifier = $this->resourceUniqueIdRetriever->retrieve($originalResource);
+
+        if ($uniqueIdentifier === null) {
+            $this->logger->info(sprintf('Cannot retrieve unique identifier for item %s', $originalResource->getUri()));
```
gabrielfs7, line 77: - [ ] This should be an error exception. If this feature flag is active why would an Item/Test be missing this id and we masking the error?

model/UniqueId/Service/UniqueIdSetter.php
```diff
@@ -0,0 +1,88 @@
...
+        if (!$this->featureFlagChecker->isEnabled('FEATURE_FLAG_UNIQUE_NUMERIC_QTI_IDENTIFIER')) {
+            $this->logger->info('Unique identifier feature not enabled.');
+
+            return $item;
+        }
+
+        $originalResourceUri = $this->resourceMetadataRetriever->getOriginalResourceUri($item);
+
+        if ($originalResourceUri === null) {
+            $this->logger->info(sprintf('Cannot retrieve original resource URI for item %s', $item->getUri()));
```
gabrielfs7, line 68: This is expected if this is not a translation, but an original item. 

We need to change the behavior here. If this is a translation and missing this metadata, throw an exception.

model/UniqueId/Service/UniqueIdSetter.php
shpran: Your translation creation service require to remove resource from post creation services

gabrielfs7 (approved):
Pre-approved as the expected behavior is working. 

Test and styling will be fixed in the integration branch

model/qti/Identifier/ServiceProvider/QtiIdentifierServiceProvider.php
```diff
@@ -0,0 +1,46 @@
...
+
+use oat\generis\model\DependencyInjection\ContainerServiceProviderInterface;
+use oat\oatbox\log\LoggerService;
+use oat\taoQtiItem\model\qti\Identifier\Service\QtiIdentifierSetter;
+use oat\taoQtiItem\model\qti\Service;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+
+use function Symfony\Component\DependencyInjection\Loader\Configurator\service;
+
+class QtiIdentifierServiceProvider implements ContainerServiceProviderInterface
```
gabrielfs7, line 33: Can we call it `QtiServiceProvider` to avoid having to create new providers later for the same namespace `qti`

PR #2611, opened 2024-11-07
chore: new logic for minMax component: min value can be 0

taopkorczak:
Related to: https://oat-sa.atlassian.net/browse/AUT-1389

Separate zero and disabled states in minMax component

Changes:
- Modified minMax component to treat 0 as a valid numeric value instead of a disabled state
- Updated disabled field behavior to use null/empty value (previously defaulted to 0)
- Fixed bug AUT-1389 by allowing lowerBound to accept 0 as a valid number

https://github.com/user-attachments/assets/9753213e-241a-4f9f-b888-d72b29708bc0


views/js/qtiCreator/widgets/component/minMax/minMax.js
```diff
@@ -47,7 +47,10 @@ define([
             //default help message
             helpMessage: __(
                 'The minimum number of choices that the candidate is required to select to form a valid response.'
-            )
+            ),
+
+            // determines if field can be null
+            canBeNull: false,
```
olga-kulish, line 53: I still don't understand, should it be possible to add a _negative_ number or not? If it was confirmed that _zero_ is minimum, could you please add a comment in the ticket?

[This comment](https://oat-sa.atlassian.net/browse/AUT-1389?focusedCommentId=284097) is wrong: in QTI https://www.imsglobal.org/question/qtiv2p1/imsqti_infov2p1.html#element10065  `lowerBound [0..1]` means that this property may or may not exist. It doesn't mean that minimum value is 0 and maximum value is 1. On the other hand, from [Dragos's comment](https://oat-sa.atlassian.net/browse/AUT-1389?focusedCommentId=284097), negative numbers _should_ be possible. His explanation seems logical.

views/js/qtiCreator/widgets/component/minMax/minMax.js
```diff
@@ -47,7 +47,10 @@ define([
...
         },
```
olga-kulish, line 54: 1. From ticket ACs, "by default, min is set to 0 and max should not be set.". Create new interaction, enable "map response", set scores for choices. `min` field in UI is unchedked+disabled and empty, but shouldn't it be checked+enabled and have `0` inside?
2. From ticket ACs, "_max should be strictly greater than min._" Now I can enter equal values: `min:18, max:18`. Is it as expected?

Current behavior is fine for me, just it contradicts ACs.

views/js/qtiCreator/widgets/component/minMax/minMax.js
olga-kulish: Also, ACs say to enter any `float` number. 

Should we ignore that and continue with integers, or should we just have a _plain text-input_ field instead of _stepper_? Like when you add a score for choice:
Screenshot_22

wazelin: Floating-point values should be allowed, that's for sure.
I will check with Dragos about the negative values. I don't think we need them, though.

views/js/qtiCreator/widgets/component/minMax/minMax.js
wazelin: Point `#1` should be addressed, yes.

views/js/qtiCreator/widgets/component/minMax/minMax.js
```diff
@@ -47,7 +47,10 @@ define([
             //default help message
```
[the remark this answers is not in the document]
olga-kulish: Still failing

views/js/qtiCreator/widgets/helpers/formElement.js
```diff
@@ -150,7 +150,7 @@ define([
         const numericValue = options.floatVal ? parseFloat(value) : parseInt(value, 10);
         let isActualNumber = !isNaN(numericValue);
 
-        if (!options.allowNull && (numericValue === 0 || !isActualNumber)) {
+        if (!options.allowNull && !isActualNumber) {
```
olga-kulish, line 153: Add OrderInteraction. In side panel, switch radio-button to _"sort from source to target list_" . Preview --> there's `minChoices: 0` in `getItem` response. In previous version, there was neither `minChoices` property.

It should be fine, because `minChoices=0` and no defined `minChoices` have the same meaning, but still, if it's easy to keep previous behavior, maybe you should. If not, let's ignore this.

views/js/qtiCreator/widgets/component/minMax/minMax.js
```diff
@@ -492,11 +546,7 @@ define([
 
                         fieldControl.input.on('change', function () {
                             self.syncValues(field);
-
-                            if (!config.allowDecimal) {
-                                self.convertToNumber(field);
-                            }
-
+                            self.convertToNumber(field);
```
olga-kulish, line 549: Now I can set negative number for `min`, but not for `max`. 

Why? I think `max` should be allowed to be negative too. If user can set negative score for interaction choice, then both `min` and `max` should be allowed to be negative too, because they limit that choice-score.

PS: Maybe it would have been better to reuse same input component which is used for choice-score? And leave `minMax.js` untouched. There would be also more space, now if you enter negative/float number, user can't even see whole number because input field is too small. But up to you.

Screenshot_23

olga-kulish (commented):
Good job! 
It seems to behave as expected (still, please check comments), and there seems to be no regressions with this UI component (though I checked only a couple of interactions).

In published test, score is calculated, and seems like it's correctly limited to `upperBound/lowerBound` (with negative and or/float numbers)

views/js/qtiCreator/widgets/component/minMax/minMax.js
taopkorczak: I did `max` value as 0 or positive value in purpose, because of acceptance criteria: [AUT-1389](https://oat-sa.atlassian.net/browse/AUT-1389)

AC:
  a. This includes setting the:
    i. min  to 0, negative, or positive double-precision numbers
    ii. max to 0, or positive double-precision numbers

olga-kulish: Oh, ok, I didn't notice  that AC was updated.

olga-kulish (approved):
Works well, good job!

PR #2612, opened 2024-11-07
feature: Remove LabelBasedLomOntologyClassificationExtractor.php

Karol-Stelmaczonek:
Ticket: https://oat-sa.atlassian.net/browse/AUT-3784 and https://oat-sa.atlassian.net/browse/AUT-3953

## What's Changed

-  Remove `LabelBasedLomOntologyClassificationExtractor.php` and additional files
- Add migration to remove it from `tao/config/taoQtiItem/metadataService.conf.php` if exists


tikhanovichA (approved):
LGTM

migrations/Version202411071618031101_taoQtiItem.php
```diff
@@ -6,27 +6,27 @@
...
     {
-        return '';
+        return 'Remove oat\taoQtiItem\model\qti\metadata\ontology\LabelBasedLomOntologyClassificationExtractor' .
+            ' if present';
     }
 
     public function up(Schema $schema): void
     {
-        $this->runAction(new AddLabelInjectorForExport());
+        $this->runAction(new RemoveLabelInjectorFromExport());
```
gabrielfs7, line 26: We are renaming the class and keeping the same migration, it will break previous installation using the old injector

tikhanovichA: Version202411071618031101_taoQtiItem migration name also new

scripts/install/ReplaceMetadataExtractor.php
```diff
@@ -1,48 +0,0 @@
-<?php
```
gabrielfs7, line 1: The 2 removed classes exist since April, are we sure this is safe to remove. 

Summoning @bartlomiejmarszal since he introduced this. 

We need to make sure this is not creating a breaking change that can affect next exports/imports

tikhanovichA: Yes, we will discuss it with Bart as we didnt find any reasons to have this extractor

migrations/Version202411071618031101_taoQtiItem.php
Karol-Stelmaczonek: IDK why its shows it as rename. Its new file generated via migration tool.

PR #2625, opened 2024-11-18
REL-1884: Delivery cannot be published when test contains Item with rich passages

vbyndych:
Related to: https://oat-sa.atlassian.net/browse/REL-1884

Steps to reproduce:
1. Go to "Deliveries" page
2. Click "New delivery" button below the Deliveries tree
3. Enter the title of Test from preconditions in the "Select the test" field (3 or more characters)
4. Click "Publish" button

The issue was connected to the interface change in the `\League\Flysystem\ProxyArrayAccessToProperties::offsetSet`. In the new version, it doesn't allow you to set new or change an existing property of the object and throws an exception in case you try to do it.


yaraslau-kavaliou (approved):
LGTM

PR #2640, opened 2024-12-02
chore: vertical direction of writing in Items with Choice Interactions

taopkorczak:
Related to: https://oat-sa.atlassian.net/browse/TR-6379

# Vertical direction of writing in Items with Choice Interactions
All the AC are applicable to items with

Choice interaction options are rendered in columns instead of rows from right to left
Both the prompt and the options text is written vertically
Both the prompt and the options text supports Ruby tags

Screenshot 2024-12-02 at 09 46 51


olga-kulish (commented):
Authoring/CurrentGen is not required yet. So for now we ignore or close this PR.

However, in the future it will almost certainly be required, then we can return to this.

PR #2655, opened 2025-01-09
fix: ruby tags focus and navigation

viktar-dzmitryieu-tao:
Ticket: https://oat-sa.atlassian.net/browse/INF-279

## What's Changed

- Updated styling for the ruby tags appeared specifically for the Chrome (v. >=128)

## How to test

- Add any interaction in test item
- Define prompt to it
- Insert Furigana (Ruby) to any selection
- See the `<rt>` content is aligned to the beginning of the element

**Deployed** to https://test-viktar-authoring.v3.playground.taocloud.org/tao/Main/login


tikhanovichA (approved):

views/scss/qti/_all-interactions.scss
```diff
@@ -10,6 +10,13 @@
     .instruction-container {
         clear: both;
     }
+
+    ruby rt {
+        // Chrome >=128 and Firefox support
+        ruby-align: start;
```
olga-kulish, line 16: I don't think we should do it. I understand that while you are typing ruby, cursor position is weird, but:
Screenshot_62


Or at least do not apply _ruby-align_ in test-runner (see Item Preview in the playground). These customers still use _CurrentGen test-runner_.

Also, if you decide to keep this, could you please add this style for _A-block_ too, not only for _Prompt_?

olga-kulish: Btw, I tried to use `range.moveToElementEditablePosition(rtElement, true)`: https://github.com/oat-sa/ckeditor-dev/pull/51 . It doesn't cause linebreak issues. However, placeholder `&nbsp;` is never removed; and if you edit furigana and delete all letters, you get an uneditable empty `<rt></rt>`. Your solution behaves better in those cases.

viktar-dzmitryieu-tao: I'll close this PR as it was not required and can lead to misunderstanding

PR #2656, opened 2025-01-09
fix: pci filename is used as identifier, not path

augustas:
Fixes a regression in PCI listing. 

PCI filename should be used as identifier, not full path. 

  
#### How to test
 
- Update local TAO to 2025.01 community version
- Navigate to PCI list
- Witness that all PCIs are present as expected

### DIFF
--- a/model/portableElement/storage/PortableElementRegistry.php
+++ b/model/portableElement/storage/PortableElementRegistry.php
@@ -145,7 +145,7 @@ abstract class PortableElementRegistry implements ServiceLocatorAwareInterface
 
         foreach ($contents as $file) {
             if ($file['type'] === 'file') {
-                $identifier = $file['path'];
+                $identifier = basename($file['path']);
                 $elements[$identifier] = $this->get($identifier);
             }
         }


model/portableElement/storage/PortableElementRegistry.php
```diff
@@ -145,7 +145,7 @@ private function getAll()
 
         foreach ($contents as $file) {
             if ($file['type'] === 'file') {
-                $identifier = $file['path'];
+                $identifier = basename($file['path']);
```
progys, line 148: test would be nice :P

oatymart (approved):
It works, PCIs are back.

model/portableElement/storage/PortableElementRegistry.php
augustas: Yeah I agree, this class is tested by integration test, but we don't run them automatically and they are neglected and failing. For this particular case `vendor/bin/phpunit taoQtiItem/test/integration/portableElement/PortableElementServiceTest.php --bootstrap generis/test/bootstrap.php` would have shown this issue

PR #2660, opened 2025-01-10
Feature/TR-6370/Item creator lang attribute and event

oatymart:
https://oat-sa.atlassian.net/browse/TR-6370

Goes with https://github.com/oat-sa/extension-tao-deliver-pcis/pull/5

Setting temporary `lang` attribute here allows AudioRecording interaction to know what item `xml:lang` value is configured.

Screenshot 2025-01-10 at 16 23 39


olga-kulish (approved):
Works nicely

PR #2666, opened 2025-01-27
fix: Prevent original test media autoplay

Nevermind23:
## Task
https://oat-sa.atlassian.net/browse/ADF-1861

## Changelog
- Prevent original test media outplay while editing translation


views/js/qtiCreator/plugins/panel/itemViewer.js
```diff
@@ -141,11 +141,17 @@ define([
...
                 viewer.on('ready', () => {
-                    // Force to restore the context after rendering the item.
-                    // This is needed as the editor needs to access the context of the item being edited.
-                    // Otherwise, the editor might not work properly.
-                    viewer.getRunner().on('renderitem', restoreCreatorContext);
-                    restoreCreatorContext();
+                    const preventAutoPlay = () => {
+                        document.querySelectorAll('audio, video').forEach(media => {
+                            media.pause();
+                        });
```
jsconan, line 147: **thought:** I'm not sure about this change. Is it about any media, or is it for media interaction? I think it should rather go to the media interaction. In the current state, it leaves me the taste of a type of coupling. Pausing the media in such a direct way may also impact the media player if any.

**suggestion:** I'd rather advise to work at the level of the media interaction.

jsconan (commented):
I'm not sure about the change. I feel incomfortable with it.

views/js/qtiCreator/plugins/panel/itemViewer.js
jsconan: **suggestion:** It would be better to scope the selector so that it only applies to the viewer.

Here are a few examples (I did not test them):

Using jQuery
```suggestion
                        $viewerContainer.find('audio, video').each((i, media) => {
                            media.pause();
                        });
```

Or vanilla:
```suggestion
                        $viewerContainer.get(0).querySelectorAll('audio, video').forEach(media => {
                            media.pause();
                        });
```

jsconan (approved):
Change scoped to the item viewer. It looks OK, but I did not test the behavior.
To be checked by QA.

shaveko (approved):
Looks good, the scoping is fixed

PR #2683, opened 2025-02-12
fix: keep `style` attribute intact when importing QTI

wazelin:
# [COECDP25-1009](https://oat-sa.atlassian.net/browse/COECDP25-1009)

This PR partially reverts #343 in an attempt to keep the `style` attributes intact when importing QTI.

[COECDP25-1009]: https://oat-sa.atlassian.net/browse/COECDP25-1009?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/helpers/Authoring.php
+++ b/helpers/Authoring.php
@@ -1,6 +1,6 @@
 <?php
 
-/*
+/**
  * This program is free software; you can redistribute it and/or
  * modify it under the terms of the GNU General Public License
  * as published by the Free Software Foundation; under version 2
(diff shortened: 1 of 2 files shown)


tikhanovichA (approved):

helpers/Authoring.php
```diff
@@ -155,16 +156,16 @@ public static function sanitizeQtiXml($qti)
     {
         $doc = self::loadQtiXml($qti);
 
-        $xpath = new \DOMXpath($doc);
+        $xpath = new DOMXpath($doc);
 
-        foreach ($xpath->query("//*[local-name() = 'itemBody']//*[@style]") as $elementWithStyle) {
-            $elementWithStyle->removeAttribute('style');
-        }
+//        foreach ($xpath->query("//*[local-name() = 'itemBody']//*[@style]") as $elementWithStyle) {
```
gabrielfs7, line 161: Good catch @wazelin . IF you feel we may have trouble disabling it, since it is enabled for 10y, we may add a FF for that and enable it only for the client that wants to use this:

```
if ($featureFlagChecker->isEnabled('FEATURE_FLAG_ALLOW_QTI_STYLE_ATTR')) {
...
}
```

wazelin: That restriction is very superficial: it only removes the attribute on import. We support it throughout the rest of the flow: export, publication, compilation, runtime.

Both Terre UI and Solar UI work fine.

That does not make our item bank QTI incompliant – we still support the strict QTI items.

PR #2685, opened 2025-02-13
[Fix] Remove old properties before setting new props during QTI import

shpran:
# [HKD-295](https://oat-sa.atlassian.net/browse/HKD-295)

## Changes
* Now old list property value(s) will be removed before setting new properties during QTI import

[HKD-295]: https://oat-sa.atlassian.net/browse/HKD-295?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/model/qti/metadata/ontology/MappedMetadataInjector.php
+++ b/model/qti/metadata/ontology/MappedMetadataInjector.php
@@ -98,7 +98,9 @@ class MappedMetadataInjector
 
     private function setListValue(Property $property, Resource $resource, SimpleMetadataValue $metadataValue): void
     {
+        $resource->removePropertyValues($property);
         $list = $this->listService->getListElements($property->getRange());
+
         foreach ($list as $listElement) {
             if (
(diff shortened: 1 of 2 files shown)


model/qti/metadata/ontology/MappedMetadataInjector.php
```diff
@@ -98,7 +98,9 @@ private function isInjectableProperty(array $mappedProperties, string $mappedPat
 
     private function setListValue(Property $property, Resource $resource, SimpleMetadataValue $metadataValue): void
     {
+        $resource->removePropertyValues($property);
```
gabrielfs7, line 101: Good catch!

Indeed it makes sure existing values will be removed and only the new ones will be set, but how does it prevent that the resource property does not get duplicated? 🤔

gabrielfs7 (approved):
Good job! :) 

PR #2693, opened 2025-02-20
Fix/INF-289/textEntry integer parse error validation 

olga-kulish:
Related to https://oat-sa.atlassian.net/browse/INF-289

- [ ] apply https://github.com/oat-sa/tao-item-runner-qti-fe/pull/426


oatymart (approved):
LGTM

PR #2728, opened 2025-04-16
fix: update error message when the same variable id

tikhanovichA:
## What's Changed 
- Change displayed message when the same variable ID

## How to test
- Add two interactions to an item
- Try to set Response identifier: `RESPONSE_1` for the first interaction

image


bartlomiejmarszal (approved):

views/js/qtiCreator/widgets/helpers/validators.js
```diff
@@ -73,7 +73,7 @@ define([
         //warning: simplistic implementation, allow only one unique identifier in the item no matter the element class/type
         {
             name: 'availableIdentifier',
-            message: __('This identifier must not be used by any other response or item variable.'),
```
pnal, line 76: Why do not delete the translations for the old message? Is it still used anywhere?

pnal (approved):

Karol-Stelmaczonek (approved):

views/js/qtiCreator/widgets/helpers/validators.js
tikhanovichA: it will be deleted after sync with Crowdin

PR #2744, opened 2025-05-21
[Fix] Prevent changing unique QTI ID

shpran:
## What's Changed
- Added validation to not allow changing QTI ID if `FEATURE_FLAG_UNIQUE_NUMERIC_QTI_IDENTIFIER` is set to `true`;

> Please tick the appropriate points
- [x] Ticket attached or not required
- [ ] Breaking change
- [ ] Configuration change
- [ ] Release version change

## Companion PRs
- https://github.com/oat-sa/extension-tao-testqti/pull/2612

## How to test
- create a new item and edit it
- use the browser’s inspector to enable the `identifier` field
- change its value
- click on `save`

Bug: item is saved with the new QTI identifier value.
Fix: an error that the QTI schema is invalid + the corresponding log with the error message that the QTI identifier is unique and cannot be changed.

### DIFF
--- a/model/qti/Service.php
+++ b/model/qti/Service.php
@@ -27,7 +27,7 @@ use oat\oatbox\filesystem\File;
 use oat\oatbox\filesystem\FilesystemException;
 use oat\tao\model\featureFlag\FeatureFlagChecker;
 use oat\tao\model\featureFlag\FeatureFlagCheckerInterface;
-use oat\taoQtiItem\model\qti\exception\QtiModelException;
+use oat\tao\model\TaoOntology;
 use oat\taoQtiItem\model\qti\parser\XmlToItemParser;
 use tao_helpers_Uri;
 use common_exception_FileSystemError;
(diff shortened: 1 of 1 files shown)


model/qti/Service.php
```diff
@@ -143,6 +146,14 @@ public function getXmlByRdfItem(core_kernel_classes_Resource $item, $language =
      */
     public function saveDataItemToRdfItem(Item $qtiItem, core_kernel_classes_Resource $rdfItem)
     {
+        if ($this->getFeatureFlagChecker()->isEnabled('FEATURE_FLAG_UNIQUE_NUMERIC_QTI_IDENTIFIER')) {
```
gabrielfs7, line 149: Not captured by QA, but should we also add it for tests?

gabrielfs7 (approved):

Nevermind23 (approved):

model/qti/Service.php
shpran: As I see, we do not have tests for this method at all 😓 
But I will try to cover at least this method

PR #2993, opened 2026-05-05
Feature/TR-6832/Grunt babel pretransform

oatymart:
https://oat-sa.atlassian.net/browse/TR-6832

Assumes https://github.com/oat-sa/grunt-tao-bundle/pull/75 installed, and configures the babel pre-transform.

To be validated on an environment with DEBUG_MODE: false set up, so the JS bundles are loaded.


kilatib (approved):
LGTM

PR #3003, opened 2026-05-19
Fix/aut 4551/rewrite items api fix 2

tikhanovichA:
Adds an optional import mode to overwrite existing items by label within the target class (class-uri / class-label), without changing the existing metadata guardian behavior.

- New API/query parameter: overwriteByLabelInTargetClass (default: false)
- Label-based overwrite runs only when both are true:
      - itemMustBeOverwritten=true
      - overwriteByLabelInTargetClass=true
- In that mode, import looks up an item in the destination class by the QTI item’s label (or title if label is empty) and overwrites it if found; otherwise it creates a new item in that class
- Metadata guardians (ItemMetadataGuardian) are not used for this lookup, so items in other classes with the same label are not affected
- When overwriteByLabelInTargetClass is false or omitted, behavior is unchanged: itemMustBeOverwritten still drives overwrite via metadata guardians (global LOM identifier match)

## Summary by CodeRabbit

* **New Features**
  * Added an optional overwriteByLabelInTargetClass option for QTI item import (sync and async) to allow overwriting items by matching labels in the target class; defaults to false and validated as a boolean.
  * Import flows and background import tasks now honor this option.

* **Documentation**
  * REST API spec updated to document the new optional form parameter.

* **Tests**
  * Adjusted unit tests for import service metadata guard setup.

[Review Change Stack](https://app.coderabbit.ai/change-stack/oat-sa/extension-tao-itemqti/pull/3003?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/model/qti/ImportService.php
+++ b/model/qti/ImportService.php
@@ -995,8 +995,7 @@ class ImportService extends ConfigurableService
         $itemClassUri = $itemClass->getUri();
         $matches = [];
         foreach ($instances as $instance) {
-            $resourceClass = $instance->getClass();
-            if ($resourceClass !== null && $resourceClass->getUri() === $itemClassUri) {
+            if ($instance->getParentClassId() === $itemClassUri) {
                 $matches[] = $instance;
             }
         }
--- a/model/qti/ImportService.php
+++ b/model/qti/ImportService.php
@@ -537,7 +537,8 @@ class ImportService extends ConfigurableService
                             $qtiModel,
                             $importMetadataEnabled,
                             $resourceIdentifier,
-                            $metadataValues
+                            $metadataValues,
+                            $metaMedataValues
                         )
                     );
 
@@ -961,19 +962,24 @@ class ImportService extends ConfigurableService
     /**
      * @param \qtism\data\AssessmentItem $qtiModel
      * @param array<string, array> $metadataValues
+     * @param array<string, \core_kernel_classes_Property> $mappedItemProperties
      */
     private function extractItemLabel(
         $qtiModel,
         bool $importMetadataEnabled = false,
         string $resourceIdentifier = '',
-        array $metadataValues = []
+        array $metadataValues = [],
+        array $mappedItemProperties = []
     ): string {
         if (
             $importMetadataEnabled
             && $resourceIdentifier !== ''
             && isset($metadataValues[$resourceIdentifier])
         ) {
-            $labelFromMetadata = $this->extractItemLabelFromMetadata($metadataValues[$resourceIdentifier]);
+            $labelFromMetadata = $this->extractItemLabelFromMetadata(
+                $metadataValues[$resourceIdentifier],
+                $mappedItemProperties
+            );
             if ($labelFromMetadata !== '') {
                 return $labelFromMetadata;
             }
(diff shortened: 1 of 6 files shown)


model/qti/ImportService.php
```diff
@@ -913,6 +942,44 @@ protected function rollback(
...
+            [OntologyRdfs::RDFS_LABEL => $label],
+            ['like' => false, 'recursive' => true]
+        );
+
+        if (empty($instances)) {
+            return false;
+        }
+
+        return reset($instances);
+    }
```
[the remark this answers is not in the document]
tikhanovichA: we will fix it here https://github.com/oat-sa/extension-tao-itemqti/pull/3003/commits/85e9f6c839bdef298eb0c4a737ba2c9e81d473a5
the fix will be related to update the last recent items if we have multiple

model/qti/ImportService.php
```diff
@@ -504,9 +513,36 @@ public function importQtiItem(
...
+                $qtiModel = null;
+
+                $overwriteByLabelInTargetClassEnabled = $itemMustBeOverwritten === true
+                    && $overwriteByLabelInTargetClass === true;
 
-                if ($enableMetadataGuardians === true) {
+                if (
+                    $overwriteByLabelInTargetClassEnabled
+                    || $enableMetadataGuardians === true
+                    || $enableMetadataValidators === true
```
bartlomiejmarszal, line 525: You are adding `$enableMetadataValidators` check on purpose?

model/qti/ImportService.php
```diff
@@ -504,9 +513,36 @@ public function importQtiItem(
...
+                        $this->extractItemLabel($qtiModel)
+                    );
+
+                    if ($guardian !== false) {
+                        \common_Logger::i(
+                            'Resource "' . $resourceIdentifier
+                                . '" matches an item label in the target class and will be overwritten.'
+                        );
+                        $overWriting = true;
+                    }
```
bartlomiejmarszal, line 551: Should we extract this into service? 
Why is this separate process on just this item? Are you sure we cannot perform this when we normally import item?

bartlomiejmarszal (commented):
This is only REST action change? This will not be possible using UI, correct? 
Just to confirm current behavior. When we try to import item in class were item exist we will just create duplicate?

model/qti/ImportService.php
bartlomiejmarszal: All current business logic already applied in ImportService therefore we can align with current implementation for this extra requirement.

model/qti/ImportService.php
bartlomiejmarszal: It was inserted for purpose

bartlomiejmarszal (approved):

model/qti/ImportService.php
```diff
@@ -504,9 +513,36 @@ public function importQtiItem(
...
+                    || $enableMetadataGuardians === true
+                    || $enableMetadataValidators === true
+                ) {
                     $this->getMetadataImporter()->setMetadataValues($metadataValues);
+                }
+
+                if ($overwriteByLabelInTargetClassEnabled) {
+                    $tmpQtiFile = $tmpFolder . helpers_File::urlToPath($qtiItemResource->getFile());
+                    $this->convertToQti2($tmpQtiFile);
+                    $qtiModel = $this->createQtiItemModel($tmpQtiFile);
```
pnal, line 533: `createQtiItemModel` used twice in method `importQtiItem` it's quite heavy operation including XML parsing, can we prevent it's run twice?

model/qti/ImportService.php
```diff
@@ -913,6 +952,70 @@ protected function rollback(
...
+                $matches[] = $instance;
+            }
+        }
+
+        if (empty($matches)) {
+            return false;
+        }
+
+        if (count($matches) > 1) {
+            \common_Logger::i(sprintf(
```
pnal, line 1076: I would move the log call to a place where a decision to overwrite is made, leaving this method responsible only for search and return items with specific label in a specific class, so it at least would be ready to be moved out and/or reused

model/qti/ImportService.php
```diff
@@ -504,9 +513,36 @@ public function importQtiItem(
...
+                if ($overwriteByLabelInTargetClassEnabled) {
+                    $tmpQtiFile = $tmpFolder . helpers_File::urlToPath($qtiItemResource->getFile());
+                    $this->convertToQti2($tmpQtiFile);
+                    $qtiModel = $this->createQtiItemModel($tmpQtiFile);
+                    $guardian = $this->findItemByLabelInClass(
+                        $itemClass,
+                        $this->extractItemLabel($qtiModel)
+                    );
+
+                    if ($guardian !== false) {
```
pnal, line 545: `findItemByLabelInClass` returns a list of instances matched as far as I see, `$guardian` var name checked as boolean looks misleading

pnal (changes requested):

model/qti/ImportService.php
```diff
@@ -913,6 +952,70 @@ protected function rollback(
...
+        if (count($matches) > 1) {
+            \common_Logger::i(sprintf(
+                'Multiple items with label "%s" found in class "%s". '
+                . 'The most recently created item will be overwritten.',
+                $label,
+                $itemClass->getLabel()
+            ));
+        }
+
+        return reset($matches);
```
Karol-Stelmaczonek, line 1084: Shouldn't it fail if detected more than 1 element? There is no guarantee the first element was the one which we want to use.

model/qti/ImportService.php
tikhanovichA: it will be called only once as we have check for the second place 
`if ($qtiModel === null) {
                    $this->convertToQti2($tmpQtiFile);`

model/qti/ImportService.php
tikhanovichA: it will return boolean or only one `return reset($matches);`

model/qti/ImportService.php
tikhanovichA: we need to find last updated and use only last one:
`'order' => TaoOntology::PROPERTY_UPDATED_AT,
                'orderdir' => 'DESC',`

model/qti/ImportService.php
tikhanovichA: This method returns only one element, which will be last updated in case there are more than two. The log message here just for info that it was more than two and it will be used and seen only in tao log

Karol-Stelmaczonek (approved):

pnal (approved):

PR #3014, opened 2026-05-31
fix/INV-1215: remove spacing around gaps matched

jebersole:
## Ticket:
https://oat-sa.atlassian.net/browse/INV-1215

## What's Changed
Fixes an issue where extra spacing was visible around gap match choices after insertion, in item previewer and current gen test runner.

image

## How to test
Create a gap match item with very long, continuous text. Highlight part of the text so that the wand appears, and replace that text with some choice. In the item previewer and current gen test runner, spaces should no longer appear and the word should look like the original.

## Summary by CodeRabbit

* **Style**
  * Gaps now align inline with surrounding content for smoother flow.
  * Clickable gap areas can shrink to their content for a tighter layout.
  * Border rendering tightened to display only intended edges for cleaner lines.
  * Edge and hover visuals refined for more consistent appearance and alignment.


views/scss/qti/_gap-match.scss
```diff
@@ -1,6 +1,10 @@
 @import 'inc/bootstrap';
 
 .qti-gapMatchInteraction {
+    .qti-gap {
+        display: inline-block;
```
olga-kulish, line 5: Maybe fix these if easy? If not, I can approve the current solution.

1. On hover, remove button isn't smoothly connected (_should be a smooth blue rectangle_)
2. On drag start, content shifts and relayouts a little (_because of change in border width?_)
3. Now it's quite difficult to distinguish the placed answer from the plain text. Can we keep the min-width as before? It seems to me that min-width helped with this usability issue. (_Can "display: inline-flex" help here?_)

Old:

https://github.com/user-attachments/assets/6e4f5d39-4918-4021-9f60-06b961a5e60f

New:

https://github.com/user-attachments/assets/95f47675-2e22-433b-b0a9-4123cc7d4014

olga-kulish (commented):
Issue is fixed, good job!

Before deploying to unit env, please also bundle js,  `taoQtiItem.min.js` includes this code too.

views/scss/qti/_gap-match.scss
jebersole: Good catches there! Thanks, I didn't realize about the blue box. I agree with all of that, but about the min-width, if we don't remove it, it seems like they might complain, as it could make a word unreadable:

Image

And I think it's still usable as the text is blue and underlined, so you can see what you have placed where, no?  The layout shifting...I think the only way would be to totally change everything and make a modal, unless you have some other ideas of how to keep the width the same when it isn't the same with and without the word inserted. This behavior is also arguably existing, as you can see if you try to insert a choice with very long text (larger than 45px, which isn't hard).
Also, `display: inline-block` seems better here as it follows the semantics of other elements, which aren't using `flex`. I don't see a difference in the browser, anyway.

olga-kulish: I still think you should return the min-width. They didn't ask to remove it.

olga-kulish (approved):
Nice fix!

There's still an issue that line-height increases when you start dragging, so interaction is "flickering". Would be nice to fix that, but it's not that important.

https://github.com/user-attachments/assets/88f8f697-3c84-4059-af17-8fff1d067cf4

olga-kulish (approved):
Good job!

PR #3070, opened 2026-07-17
feat: Add Scroll Functionality to Choice Interaction and Extended Tex…

jebersole:
…t Interaction

## Ticket:
https://oat-sa.atlassian.net/browse/INF-453

## What's Changed
Adds the Enable Scrolling checkbox and associated scrolling to Choice and Extended Text interactions. This gets set in authoring and then has an effect in CG previewer and test runner.
image
image

## How to test
Add either of the above interactions to authoring and click the checkbox, then save. There should now be a scrollbar on the item in the previewer and test runner. Other interactions should not have this checkbox.

## Summary by CodeRabbit

* **New Features**
  * Added a scrolling toggle control to supported choice, extended text, and static item forms.
  * Scrolling controls now automatically reflect writing mode and scrolling requirements.
  * Added contextual help for enabling or disabling scrolling.

* **Bug Fixes**
  * Improved synchronization between scrolling settings and their associated controls.
  * Corrected scrolling width configuration behavior.

* **Tests**
  * Expanded coverage for scrolling controls, visibility, writing modes, and interaction types.


views/js/test/qtiCreator/widgets/static/helpers/itemScrollingMethods/test.js
```diff
@@ -0,0 +1,262 @@
...
+            const clearedClasses = itemScrollingMethods.cutScrollClasses(`foo ${classes} bar`);
+
+            itemScrollingMethods.setIsVertical($form, true);
+
+            assert.strictEqual($form.attr('data-is-vertical'), 'true', 'vertical state is stored on the form');
+            assert.ok(classes.indexOf('key-navigation-focusable') >= 0, 'generated classes include the focus utility');
+            assert.ok(classes.indexOf('tao-overflow-y') >= 0, 'generated classes include the overflow utility');
+            assert.ok(classes.indexOf('tao-quarter-height') >= 0, 'generated classes include the mapped size utility');
+            assert.notOk(clearedClasses.indexOf('tao-overflow-y') >= 0, 'cutScrollClasses removes overflow utilities');
+            assert.notOk(clearedClasses.indexOf('tao-quarter-height') >= 0, 'cutScrollClasses removes mapped size utilities');
```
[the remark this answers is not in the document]
jebersole: I verified this finding against current code and skipped it as not still valid.
Reason:
[`itemScrollingMethods.getScrollClasses()` (line 264)](/Users/je/oat/extension-tao-itemqti/views/js/qtiCreator/widgets/static/helpers/itemScrollingMethods.js:264) always builds classes with newUIclass, which is defined as tao-overflow-y, and it maps size classes through options, not any vertical-specific width class list.
[`cutScrollClasses()` (line 258)](/Users/je/oat/extension-tao-itemqti/views/js/qtiCreator/widgets/static/helpers/itemScrollingMethods.js:258) correspondingly removes tao-overflow-y plus the same height-class set.
setIsVertical() only stores data-is-vertical on the form; getScrollClasses() does not read that state, so updating the test to expect tao-overflow-x and width-specific classes would contradict the current implementation rather than verify it.
Validation:
Reviewed current implementation and current test at:[itemScrollingMethods.js (line 216)](/Users/je/oat/extension-tao-itemqti/views/js/qtiCreator/widgets/static/helpers/itemScrollingMethods.js:216)
[itemScrollingMethods/test.js (line 194)](/Users/je/oat/extension-tao-itemqti/views/js/test/qtiCreator/widgets/static/helpers/itemScrollingMethods/test.js:194)

No code changes were made for this finding.

model/qti/interaction/ChoiceInteraction.php
oatymart, line 1: I'm not sure about those model changes. I don't really know enough about it to say definitively. But since we have numerous interactions, and all of them support any number of `data-` attributes as a fundamental part of the QTI item model, we have probably never bothered to explicitly define or enumerate those attributes on the PHP side. Additionally, I don't think adding this code actually _does_ anything, so we probably don't need it.

views/js/qtiCreator/widgets/interactions/helpers/interactionScrolling.js
```diff
@@ -0,0 +1,107 @@
...
+ * GNU General Public License for more details.
+ *
+ * You should have received a copy of the GNU General Public License
+ * along with this program; if not, write to the Free Software
+ * Foundation, Inc., 31 Milk St # 960789 Boston, MA 02196 USA.
+ *
+ * Copyright (c) 2020-2026 (original work) Open Assessment Technologies SA ;
+ */
+
+define(['taoQtiItem/qtiCreator/widgets/static/helpers/itemScrollingMethods'], function (itemScrollingMethods) {
```
oatymart: In my limited understanding, interaction scrolling re-uses some of the behaviours from item scrolling. Namely, the options list, the `tao-*` styles, but also some methods like `setIsScrolling`, `toggleScrollingSelect`.

This is confusing to me, as the call to `itemScrolling.setIsScrolling()` looks like it will set the scrolling on the item level. I wonder if it is possible to do a clean separation where `itemScrolling` manages the item, and `interactionScrolling` manages an interaction, without one importing the other. (The options list could be separated.) Would it cause too much duplication? Would it make any sense to try that?

views/js/qtiCreator/widgets/interactions/choiceInteraction/states/Question.js
oatymart, line 1: I see now that it mixes and matches what it needs from item scrolling and interaction scrolling. So maybe what my previous comment said would be too big of a change.

oatymart (commented):
Looks promising; I am still trying to understand the correct setup for an item, since some of the structures you can set up don't result in scrollable interactions.

wazelin:
I may be very wrong, but this looks like a lot more changes than I expected from this PR.
Weren't the requirements here boiling down to basically "keep the `Enable scrolling` checkbox available regardless of the vertical/horizontal layouts combination"?

jebersole:
> I may be very wrong, but this looks like a lot more changes than I expected from this PR. Weren't the requirements here boiling down to basically "keep the `Enable scrolling` checkbox available regardless of the vertical/horizontal layouts combination"?

Yes, but the checkbox on its own doesn't always have an effect. In order to show the scrollbar for these interactions, there needed to be logic to handle it - there was already a mechanism with the scrolling plugin.

views/js/qtiCreator/widgets/interactions/helpers/interactionScrolling.js
jebersole: They are quite similar, yes, though the reason being I wanted to constrain changes just to the interactions needed. Can you give me an example of what you mean?

model/qti/interaction/ChoiceInteraction.php
jebersole: @oatymart In `getUsedAttributes()` in extension-tao-itemqti/model/qti/Element.php lines 111 and 178, there is a whitelist-checking of attributes. In order for `data-scrolling-*` attributes to survive save and be written into the QTI, this needs to be here.
JS makes the checkbox appear and updates DOM/model state in authoring.
PHP makes data-scrolling and data-scrolling-height first-class saved attributes for those two interaction types.

olga-kulish:
> I may be very wrong, but this looks like a lot more changes than I expected from this PR. 

I'll check it tomorrow. Indeed, there should be less changes.

model/qti/interaction/ChoiceInteraction.php
oatymart: Ok, thanks. I guess I had added data attributes to something before without doing this, but it must have been some interaction or element for which `getUsedAttributes` was not overridden.

views/js/qtiCreator/widgets/interactions/helpers/interactionScrolling.js
oatymart: I think my concern is that with both `itemScrolling` and `interactionScrolling` existing, it's pretty hard for another developer to figure out what method to use where. Both modules are fully undocumented, so you have to check usages. `itemScrolling` has usages both in interactions, static elements, AND the `interactionScrolling`.

I'm not saying to change everything, but perhaps adding JSDoc descriptions on all the methods would be a good help.

views/js/qtiCreator/widgets/interactions/choiceInteraction/states/Question.js
```diff
@@ -394,13 +405,20 @@ define([
 
                     // draw scrolling methods
```
olga-kulish: 1. Unstable state on focus/blur of the widget - sometimes dropdown isn't shown, sometimes checkbox is disabled (but should be enabled checkbox and visible dropdown):

https://github.com/user-attachments/assets/7616d8bb-f502-46da-8ca8-2980f467994c

2. Vertical item -> add choice interaction -> swtich interaction to Horizontal -> switch interaction to Vertical -> dropdown with sizes is not shown, though it should be. Scroll classes/attributes won't be saved, even though checkbox is shown as checked.

views/js/test/qtiCreator/widgets/interactions/extendedTextInteraction/states/Question/test.js
```diff
@@ -158,4 +161,100 @@ define([
...
+                    maxWords: { label: 'Max Words', selected: false },
+                    pattern: { label: 'Pattern', selected: false }
+                },
+                constraintsAvailable: true
+            },
+            scrollingTplVars
+        );
+    }
+
+    function createChoiceTemplateContext(scrollingTplVars) {
```
olga-kulish: Please remove tests for choice/slider/associate from this file. This file is only for extendedText interaction.

PS: I don't mind if there are no new tests for new functionality (checkbox), just let's ensure existing tests don't fail.

views/js/test/qtiCreator/widgets/static/helpers/itemScrollingMethods/test.js
```diff
@@ -0,0 +1,113 @@
...
+ * Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+ */
+
+define([
+    'jquery',
+    'taoQtiItem/qtiCreator/widgets/static/helpers/itemScrollingMethods'
+], function ($, itemScrollingMethods) {
+    'use strict';
+
+    QUnit.module('qtiCreator/widgets/static/helpers/itemScrollingMethods');
```
olga-kulish: This repository doesn't have such low-level unit tests for UI, there are only more of "integration" tests on widgets themselves (`extendedTextInteraction/states/Question/test.js` and so on).

I think it's good. It's more useful to test the final result in the interaction, including how changing values in radio/dropdown influences interaction data attributes.

If you are absolutely that sure this test is needed, please leave it.

olga-kulish (commented):
Good job! Code looks good, item data is saved correctly, behavior is as required, but please check a bug described in the comment.

views/js/qtiCreator/widgets/interactions/choiceInteraction/states/Question.js
```diff
@@ -376,6 +394,7 @@ define([
                     $form.data('isItemVertical', isItemVertical);
 
                     $form.find('.writingMode-panel').toggle(isVerticalSupported);
+                    $form.find('.scrolling-toggle-container').css('display', isVerticalSupported ? 'block' : 'none');
```
kilatib, line 397: please show this toggle even when vertical writing is not supported

views/js/qtiCreator/widgets/interactions/extendedTextInteraction/states/Question.js
```diff
@@ -209,13 +209,35 @@ define([
...
+        };
+
         const toggleVerticalWritingModeByLang = (widget, $form, interaction) =>
             verticalWritingEditing
                 .checkItemWritingMode(widget)
                 .then(({ isVerticalSupported, isItemVertical }) => {
                     $form.data('isItemVertical', isItemVertical);
 
                     $form.find('.writingMode-panel').toggle(isVerticalSupported);
+                    $form.find('.scrolling-toggle-container').css('display', isVerticalSupported ? 'block' : 'none');
```
kilatib, line 240: please show this toggle even when vertical writing is not supported

kilatib (changes requested):
It is much better than the first iteration of changes, but please do not hide Enable Scrolling behind isVerticalSupported — the checkbox should stay available for Choice/Extended Text even when vertical writing is not supported (same idea as @wazelin  noted). Also please re-check @olga-kulish 's focus/blur and Vertical→Horizontal→Vertical cases after the stabilization fix,  I pretty sure it's unstable because you remove waitForElements parts; please recheck that case

views/js/qtiCreator/widgets/interactions/extendedTextInteraction/states/Question.js
```diff
@@ -231,44 +253,13 @@ define([
...
-                    const isScrolling = itemScrollingMethods.isScrolling(interaction);
+                    const isForcedScrolling = isWritingModeMismatch(isVertical ? 'vertical' : 'horizontal');
+                    syncScrollingControls({ isForcedScrolling });
                     itemScrollingMethods.setIsVertical($form, !!$form.data('isItemVertical'));
-                    itemScrollingMethods.toggleScrollingSelect($form, isScrolling);
                 });
 
         toggleVerticalWritingModeByLang(_widget, $form, interaction);
 
-        function waitForElement(selector, callback) {
```
kilatib, line 241: The elements do not apply to pages imitiadly those function was added for stability

jebersole: In these two files, waitForElement('input[name="writingMode"]:checked', ...) was only there to run a second initialization pass after the writing-mode radios existed. But by that point we already have enough state from:
the interaction classes
$form.data('isItemVertical')
persisted data-scrolling
So the extra waitForElement path was redundant, and worse, it was the thing re-overwriting the checkbox state on reopen.
Removing it is the right move here because:
the form HTML is already inserted before this code runs
toggleVerticalWritingModeByLang(...) already sets the writing-mode radios
we can compute forced scrolling directly from model/UI state without waiting on a checked radio selector
So yes: for this checkbox initialization, we do not need waitForElement at all.

jebersole:
> It is much better than the first iteration of changes, but please do not hide Enable Scrolling behind isVerticalSupported — the checkbox should stay available for Choice/Extended Text even when vertical writing is not supported (same idea as @wazelin noted). Also please re-check @olga-kulish 's focus/blur and Vertical→Horizontal→Vertical cases after the stabilization fix, I pretty sure it's unstable because you remove waitForElements parts; please recheck that case

@kilatib I am actually following Olga's requirements here - @olga-kulish can you confirm?

views/js/qtiCreator/widgets/interactions/extendedTextInteraction/states/Question.js
kilatib: For reproduce unstable behaviour, put at one test item: 
- multiple choice 
- A block 
- text interaction 

Then change the vertical position of the item and sub-items by switching between them; then state that "Direction of wiring" should always be saved after each change and not be lost after switching to another item in a section. 
It seems to me now it is broken

kilatib: When I edit A block position, it resets other items as it should when I change an item position. 
@olga-kulish Maybe it is ok? 

But now we should show the choices all the time, so I think AC is required 

https://github.com/user-attachments/assets/0ec68511-3918-4808-82bb-e865430b4502

kilatib (approved):
- [x] Commits are following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/)
- [x] Changelog is updated according to changes (if applicable)

views/js/qtiCreator/widgets/interactions/extendedTextInteraction/states/Question.js
jebersole: It looks like this is about all of the interactions' Direction of Writing getting reset when you click on a different Direction of Writing for the item. Confusing yes, but it appears to be existing behavior, e.g. in https://github.com/oat-sa/extension-tao-itemqti/blob/develop/views/js/qtiCreator/widgets/interactions/choiceInteraction/Widget.js#L47

olga-kulish (approved):
Works as requested, good job!

PS: for the future, please don't update bundles in the same commit with meaningful files, as in https://github.com/oat-sa/extension-tao-itemqti/pull/3070/changes/ec74b22876926b79a471bc9b34ad05c742f4100e. Bundle should have its own commit `chore: bundle`.

